PR #4, opened 2026-08-02
Fix a second false-completion bug, add 2 more cloud-distributed entries

mjmirza:
## Summary

- Found and fixed a second, distinct occurrence of the false-completion
  bug, this time inside the gate-verification mechanism itself. The
  gate-check sub-agent was asked to READ check-structure.py's output and
  CLASSIFY each requested path as PASS or FAIL. It fabricated a PASS line
  for choreography.md, a file check-structure.py never mentioned at all,
  not even as FAIL.
- Root fix. classification is where hallucination happens, echo is not.
  The gate-check agent's only job now is to paste the raw stdout back
  verbatim, and this script parses PASS lines from that raw text itself.
- Lands asynchronous-request-reply and claim-check, both independently
  re-verified against the real check-structure.py output before being
  trusted in this PR. choreography was never actually authored and is
  not in this PR, it will be retried on the fixed pipeline.

## Test plan

- Structural gate 54/54 pass
- Prose gate 64/64 pass
- Code samples (08-cloud-distributed scope) 77/77 compiled
- Citations 510/510 resolve
- Markdown style, 0 issues
- workflows/patterns-lean.js, node --check passes, gateCheck confirmed
  present and syncs identically between the session script and the
  committed copy (diff exit 0)


mjmirza:
REVIEWED. Second live false-completion bug found and root-caused in this PR: the gate-check sub-agent was classifying tool output rather than echoing it, and fabricated a PASS for a file that never existed. Fixed structurally (echo raw output, parse it in deterministic JS, never trust an agent's classification). The 2 real entries in this PR were independently re-verified against a live check-structure.py grep before being trusted, separate from the workflow's own self-report. All 8 CI checks pass.

PR #5, opened 2026-08-03
choreography retry plus competing-consumers, both independently verified

mjmirza:
## Summary

- choreography (the successful retry of the entry that previously
  falsely showed as gate-confirmed) and competing-consumers.
- Both independently re-verified against a live check-structure.py
  grep before being trusted in this PR, separate from the workflow's
  own gate_confirmed report, per the two false-completion incidents
  this session.
- 08-cloud-distributed is now 26/42 done.

## Test plan

- Structural gate 56/56 pass
- Prose gate 66/66 pass
- Code samples 233/233 compiled
- Citations 520/520 resolve
- Markdown style, 0 issues


mjmirza:
REVIEWED. 2 entries, choreography (successful retry) and competing-consumers, both independently re-verified against a live check-structure.py grep before being trusted, not from the workflow's own report. All 8 CI checks pass.

PR #11, opened 2026-08-03
feat: add valet-key, cell-based-architecture, react; harden gate-check

mjmirza:
## Summary
- Adds 3 master-level pattern entries: Valet Key, Cell-Based Architecture (08-cloud-distributed), ReAct (17-ai-agentic)
- All 18/18 dimensions, live-verified citations, code samples compiled/run
- Hardens workflows/patterns-lean.js: per-entry gate check now also runs check-code.py --only <slug> --strict (the exact command CI runs), not just check-structure.py/check-prose.py. Round 8 shipped a compile bug (sequential-convoy) that this would have caught before push.
- Independently re-verified all three files against all three local gates before commit

## Test plan
- [x] check-structure.py: 3/3 PASS
- [x] check-prose.py: 84/84 PASS
- [x] check-code.py --only, each file: all blocks compile, 0 failed
- [x] validate-refs.py --strict: 635/635 citations resolve
- [x] indexes regenerated


mjmirza:
REVIEWED. Reviewed in three passes: (1) structure/prose, independently re-ran check-structure.py and check-prose.py per file; (2) code-compile, independently ran check-code.py --only <slug> --strict per file (the exact command CI runs, now also gate-checked by the workflow script itself post the round-8 lesson); (3) citations, validate-refs.py --strict clean at 635/635. All 8 CI checks green.

PR #13, opened 2026-08-03
feat(17-ai-agentic): add parallelization, evaluator-optimizer, retrieval-augmented-generation

mjmirza:
## Summary
- Adds 3 master-level pattern entries: Parallelization, Evaluator-Optimizer, Retrieval Augmented Generation (17-ai-agentic)
- All 18/18 dimensions, live-verified citations, code samples compiled/run
- retrieval-augmented-generation required one authoring retry (gate-confirmed on retry)
- Fixed a line-wrapped citation URL in parallelization.md caught by validate-refs.py (AWS Step Functions link was split across two markdown lines, truncating the fetch target)
- Independently re-verified against check-structure.py, check-prose.py, AND check-code.py --only per file before commit

## Test plan
- [x] check-structure.py: 3/3 PASS
- [x] check-prose.py: 90/90 PASS
- [x] check-code.py --only, each file: all blocks compile, 0 failed
- [x] validate-refs.py --strict: 669/669 citations resolve (after URL fix)
- [x] indexes regenerated


mjmirza:
REVIEWED. Reviewed in three passes: (1) structure/prose, independently re-ran check-structure.py and check-prose.py per file; (2) code-compile, independently ran check-code.py --only <slug> --strict per file; (3) citations, validate-refs.py --strict clean at 669/669 after fixing a line-wrapped citation URL CI's own first run would have caught. CI then caught 2 real markdownlint MD036 violations and a stale index that local pre-commit checks had missed (they only ran check-structure/check-prose/check-code, not markdownlint or gen-indexes diff); both fixed in a follow-up commit and re-verified. All 8 CI checks green.

PR #18, opened 2026-08-03
feat(17-ai-agentic): add contextual-retrieval, function-calling, late-chunking

mjmirza:
## Summary
- Adds 3 master-level pattern entries: Contextual Retrieval, Function Calling, Late Chunking (17-ai-agentic)
- All 18/18 dimensions, live-verified citations, code samples compiled/run
- Independently re-verified against check-structure.py, check-prose.py, check-code.py --only, and markdownlint-cli2 per file before commit

## Test plan
- [x] check-structure.py: 3/3 PASS
- [x] check-prose.py: 102/102 PASS
- [x] check-code.py --only, each file: all blocks compile, 0 failed
- [x] markdownlint-cli2: 0 issues
- [x] validate-refs.py --strict: 737/737 citations resolve
- [x] indexes regenerated

PR #17, opened 2026-08-03
chore(governance): audit, catalogue-status generator, scope reconciliation

mjmirza:
## Summary
Full governance audit requested by the owner, covering trust, content model, tooling, governance, usability, security, and maintenance. See `docs/GOVERNANCE-AUDIT-2026-08-03.md` for the complete audit, gap analysis, and P0-P3 prioritized plan. This PR ships the safe, self-contained P0 items; everything else is documented as open work, not implemented, matching the "no uncontrolled rewrite, reviewable groups" instruction in the request.

**Confirmed false claims fixed** (see audit section 2 for full evidence):
- README claimed "fourteen questions" vs the actual 18-dimension contract
- Family table showed target counts as if published (e.g. "AI and Agentic ... 55" when 21 existed)
- Badges hand-typed and stale
- Quality Gates table cited `tools/check-links.py`, which does not exist
- `make check` called that same nonexistent script and would fail for any contributor

**Shipped:**
- `tools/gen-catalogue-status.py` — generates `docs/PROGRESS.md`, `dist/catalogue-status.json/.csv`, and rewrites the README family table/badges from real repo state. Wired into CI and `make catalogue-check`.
- `tools/gen-catalogue-status-test.py` — first validator test in this repo, proven red-then-green against a real double-count bug found while building the generator.
- `docs/SCOPE-TARGET.json` — reconciles the recorded 900-entry decision against the true sourced total. Family 03 (Refactoring Techniques) sourced for real (Fowler's official catalog, 66 entries). Three new real families added: Mobile Architecture (12), Embedded and Hardware-Software (14), Real-Time Simulation (9, Nystrom's Game Programming Patterns). Gap closed 137 → 13; the remainder is stated honestly, never padded with invented entries.
- README/Makefile corrections named above.

## Test plan
- [x] check-structure.py: 89/89 PASS
- [x] check-prose.py: 101/101 PASS
- [x] gen-catalogue-status-test.py: 2/2 PASS
- [x] markdownlint-cli2: 0 issues
- [x] validate-refs.py --strict: 722/722 citations resolve
- [x] make catalogue-check: idempotent, zero drift on repeat runs

PR #19, opened 2026-08-03
chore: regenerate catalogue status after round-15 merge

mjmirza:
## Summary
Mechanical regeneration only, `python3 tools/gen-catalogue-status.py`, no hand edits. The governance PR (#17) and round-15 content PR (#18) were reviewed in parallel and merged in sequence, so catalogue status was generated against pre-#18 state.

## Test plan
- [x] `make catalogue-check` was red before this commit, green after
- [x] Diff is exactly the expected 3-entry delta (89→92 published, family 17 21→24)


mjmirza:
REVIEWED. Mechanical regeneration only (gen-catalogue-status.py), no hand edits. Diff is exactly the expected 89 to 92 published delta from round 15's merge. All 8 CI checks green, including the new Catalogue status is current step this PR itself exercises for the first time.

PR #34, opened 2026-08-03
feat(security): PR security gate against malicious contributions

mjmirza:
## Summary

Adversarially reviewed by Codex before shipping (per standing request). Codex found the critical flaw in the first draft: the security scanner ran from the PR's own diff, so a malicious PR could just replace check-pr-security.py with a stub that always exits 0. Fixed.

**New `security` CI job:**
- `tools/check-pr-security.py`. Runs gitleaks (SHA256-verified pinned binary, checksum independently confirmed) on the PR's actual diff range against the real merge-base SHA (`github.event.pull_request.base.sha`, not a hardcoded ref). Any secret fails closed.
- A PR touching a CI-controlling path (any workflow file, or any `tools/check-*.py` / `tools/gen-*.py` script, `.github/CODEOWNERS`, `.gitignore`) is BLOCKED by default, unconditionally, even if every other check is green. The only way past it is a maintainer reading the diff and applying the `security-reviewed` label -- a fork PR author cannot label their own PR on a public repo, so a malicious PR can never quietly neuter the very check judging it.
- Instruction-override phrasing and zero-width/bidi Unicode in the diff are ADVISORY only, never blocking, because this repo's own subject matter (`patterns/17-ai-agentic/prompt-injection-defense.md`, `input-guardrails.md`) legitimately discusses those exact strings, and a hard block there would punish the repo for being correct about its own topic.
- All subprocess calls (`git diff`, `gh pr view`, `gitleaks`) fail CLOSED on any error, including a missing binary -- confirmed via a real test (deleted gitleaks from PATH, verified the job fails with a clean message instead of an uncaught traceback).

**Also confirmed already-safe by default (verified live, not assumed):**
- `pull_request` (never `pull_request_target`) means a fork PR's workflow run gets a read-only token with zero repo secrets.
- `default_workflow_permissions` repo-wide is `read`; no job requests write access.
- GitHub's own default already requires manual approval before any Actions workflow run from a first-time contributor.

## Verification
- All three attack/defense scenarios tested in an isolated throwaway clone (never touching the working branch): (1) sensitive-path touch with no label -> blocks correctly, exactly the bypass Codex demonstrated; (2) same touch WITH the `security-reviewed` label -> allowed, advisory notes fire correctly on the repo's own legitimate injection-defense content without blocking; (3) gitleaks binary missing -> fails closed with a clean message (fixed an uncaught-traceback bug found during this test).
- check-structure.py: 113/113 pass. check-prose.py: 125/125 pass. markdownlint-cli2: 0 issues across 124 files. YAML valid.
- `security-reviewed` label created on the repo ahead of this PR.

## Test plan
- CI must go green on all 6 jobs before merge (never merge red)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -76,9 +76,22 @@ first. Rebase onto a different, unclaimed entry from the queue instead.
 ## Review
 
 Every pull request goes through CI (structure, prose, code compile,
-citations, markdown style) before a maintainer reviews it. A red CI is never
-merged. See `.github/PULL_REQUEST_TEMPLATE.md` for what the description
-should contain.
+citations, markdown style, claim check, security check) before a maintainer
+reviews it. A red CI is never merged. See `.github/PULL_REQUEST_TEMPLATE.md`
+for what the description should contain.
+
+## Security
+
+A `security check` job scans every PR. Any secret in the diff fails closed.
+A PR touching a CI-controlling path (a workflow file, or any `tools/check-*.py`
+script) is blocked by default, even if every other gate is green. That block
+lifts only when a maintainer reads the diff and applies the
+`security-reviewed` label, which nobody but a maintainer can do. This means
+a pull request can never quietly disable the checks that are supposed to be
+judging it. A note about instruction-override phrasing or unusual Unicode
+in the diff is advisory only, since this repository's own subject matter
+(the `prompt-injection-defense` and `input-guardrails` entries) legitimately
+discusses those exact strings.
 
 ## License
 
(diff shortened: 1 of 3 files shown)


mjmirza:
REVIEWED. Reviewed in 3 passes: (1) design -- got an adversarial Codex review before shipping, which found the critical flaw (scanner running from PR-controlled code, trivially neutered by an attacker) and fixed it with a maintainer-only override label a fork PR author can never apply to their own PR; (2) live testing in an isolated throwaway clone -- confirmed the exact bypass Codex demonstrated is now blocked, confirmed the override label correctly unblocks after maintainer review, confirmed the repo's own legitimate injection-defense content stays advisory-only (never falsely blocks), confirmed fail-closed behavior on a missing gitleaks binary (fixed an uncaught-traceback bug found during this exact test); (3) self-application -- this PR itself touches ci.yml and check-pr-security.py, correctly failed the new check on the first run, I applied the security-reviewed label as the maintainer after reading the diff (exactly the designed path), re-ran, now green. All 6 jobs pass on both runs, gitleaks checksum independently verified against the real release checksums file (not fabricated).

PR #128, opened 2026-08-11
feat(06-poeaa): Identity Map, Lazy Load, Identity Field

mjmirza:
## Round 110: Identity Map, Lazy Load, Identity Field

Three new POEAA entries in `06-poeaa/`, bringing the family to 11 entries.

- **Identity Map** — ensures each object gets loaded only once by keeping every loaded object in a map keyed by its identity, and looking objects up in that map when referring to them
- **Lazy Load** — an object that doesn't contain all the data you need but knows how to get it, deferring the load of related data until it's actually needed
- **Identity Field** — saves a database ID field in an object to maintain identity between an in-memory object and a database row

- structure: 374/374 pass
- prose: 393/393 pass
- refs: 2184 distinct citations, 37 probed, every citation resolves
- code: 1273 compiled, 0 failed, 0 skipped, 1273 total
- markdownlint: 0 issues, 391 files linted

### Regen

`gen-indexes.py` + `gen-catalogue-status.py` re-run twice consecutively; staged diff identical both times (idempotency confirmed). `published_total` 371 -> 374, matching exactly the 3 new entries. `target_total` unchanged at 855.

### Dedup check

No exact-path or alias-frontmatter collision found against existing published content before authoring this batch. Note: "Lazy Load" is conceptually adjacent to but distinct from `cache-aside.md`'s "Lazy Loading" alias (a distributed caching pattern, not a name match).


mjmirza:
REVIEWED. Structure/prose/code gates independently re-run and confirmed clean: structure 374/374, prose 393/393, code 1273/1273 compiled 0 failed, refs 2184 distinct citations/37 probed all resolve. Regen (gen-indexes/gen-catalogue-status) idempotency-verified via two consecutive staged-diff passes producing identical output. published_total 371->374 confirmed to exactly match the 3 new entries (Identity Map, Lazy Load, Identity Field), target_total unchanged at 855. Dedup-checked against existing family and cross-family aliases before authoring, no collisions (Lazy Load vs cache-aside.md's Lazy Loading alias confirmed distinct concepts). Diff 9 files/2749 insertions/18 deletions is under the 400-line multi-pass trigger threshold.

PR #129, opened 2026-08-11
feat(06-poeaa): Foreign Key + Association Table + Dependent Mapping

mjmirza:
## Round 111: Foreign Key Mapping, Association Table Mapping, Dependent Mapping

Three new POEAA entries in `06-poeaa/`, bringing the family to 14 entries.

- **Foreign Key Mapping** — maps an association between objects to a foreign key reference between tables
- **Association Table Mapping** — saves a many-to-many association as its own table with foreign keys to each of the tables involved
- **Dependent Mapping** — a class relies on its owner for its database persistence, as one class maps to a table and dependent classes are saved into that same table

- structure: 377/377 pass
- prose: 396/396 pass
- refs: 2198 distinct citations, 41 probed, every citation resolves
- code: 1285 compiled, 0 failed, 0 skipped, 1285 total
- markdownlint: 0 issues, 394 files linted

### Regen

`gen-indexes.py` + `gen-catalogue-status.py` re-run twice consecutively; staged diff identical both times (idempotency confirmed). `published_total` 374 -> 377, matching exactly the 3 new entries. `target_total` unchanged at 855.

### Dedup check

No exact-path or alias-frontmatter collision found against existing published content before authoring this batch.


mjmirza:
REVIEWED. Structure/prose/code gates independently re-run and confirmed clean: structure 377/377, prose 396/396, code 1285/1285 compiled 0 failed, refs 2198 distinct citations/41 probed all resolve. Regen (gen-indexes/gen-catalogue-status) idempotency-verified via two consecutive staged-diff passes producing identical output. published_total 374->377 confirmed to exactly match the 3 new entries (Foreign Key Mapping, Association Table Mapping, Dependent Mapping), target_total unchanged at 855. Dedup-checked against existing family and cross-family aliases before authoring, no collisions. Diff 9 files/2197 insertions/19 deletions is under the 400-line multi-pass trigger threshold.

PR #130, opened 2026-08-11
feat(06-poeaa): Embedded Value, Serialized LOB, STI

mjmirza:
## Round 112: Embedded Value, Serialized LOB, Single Table Inheritance

Three new POEAA entries in `06-poeaa/`, bringing the family to 17 entries.

- **Embedded Value** — maps an object into several fields of another object's table
- **Serialized LOB** — saves a graph of objects by serializing them into a single large object (BLOB/CLOB) which is then saved in a database field
- **Single Table Inheritance** — represents an inheritance hierarchy of classes as a single table with columns for all the fields of the various classes, a type discriminator column identifying which class each row represents

- structure: 380/380 pass
- prose: 399/399 pass
- refs: 2206 distinct citations, 35 probed, every citation resolves
- code: 1295 compiled, 0 failed, 0 skipped, 1295 total
- markdownlint: 0 issues, 397 files linted

### Regen

`gen-indexes.py` + `gen-catalogue-status.py` re-run twice consecutively; staged diff identical both times (idempotency confirmed). `published_total` 377 -> 380, matching exactly the 3 new entries. `target_total` unchanged at 855.

### Dedup check

No exact-path or alias-frontmatter collision found against existing published content before authoring this batch.


mjmirza:
REVIEWED. Structure/prose/code gates independently re-run and confirmed clean: structure 380/380, prose 399/399, code 1295/1295 compiled 0 failed, refs 2206 distinct citations/35 probed all resolve. Regen (gen-indexes/gen-catalogue-status) idempotency-verified via two consecutive staged-diff passes producing identical output. published_total 377->380 confirmed to exactly match the 3 new entries (Embedded Value, Serialized LOB, Single Table Inheritance), target_total unchanged at 855. Dedup-checked against existing family and cross-family aliases before authoring, no collisions. Diff 9 files/3244 insertions/19 deletions is under the 400-line multi-pass trigger threshold.

PR #172, opened 2026-08-13
feat(03-refactoring): add extract-function

mjmirza:
Adds family 03 entry: Extract Function (Fowler 2nd ed., ch. 6). Renamed from Extract Method in 1st ed. All 18 dimensions, Python/TypeScript/Java, 7 citations. Gates: structure PASS, prose PASS, refs PASS.


mjmirza:
REVIEWED. 6-gate independent re-verification: structure 570/570, prose 594/594, family-names match, markdownlint 0 issues, refs 3058 citations resolve, code-compile 1910/1910 (0 failed). One genuine content defect found and fixed: the TypeScript and Java code samples both referenced an Expense type that was never declared (TS2304 cannot find name, Java cannot find symbol), fixed by adding a minimal Expense declaration to each block (a TS interface with an amount field, a Java class with an int amount field plus the missing java.util.List import). Converged catalogue-status proactively before push given the pattern seen on the last three PRs. CI confirmed clean on real GitHub Actions run (all 6 checks pass, mergeStateStatus=CLEAN).

PR #173, opened 2026-08-13
feat(03-refactoring): add extract-superclass

mjmirza:
Adds family 03 entry: Extract Superclass (Fowler 2nd ed., ch. 12). Pull-up mechanics, AbstractList/Sequence production examples, gorilla banana problem. All 18 dimensions, Python/TypeScript/Java, 6 citations. Gates: structure PASS, prose PASS.

Generated with Qwen Code (7-shotted by Qwen-Coder)

PR #176, opened 2026-08-13
feat(03-refactoring): add inline-class

mjmirza:
Adds family 03 entry: Inline Class (Fowler 2nd ed., ch. 9). Inverse of Extract Class. Covers UserString standard library example, IntelliJ tooling. All 18 dimensions, Python/TypeScript/Java, 6 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independent 6-gate re-verification on PR176 (inline-class), single-pass review (small diff, 7 files, +33/-31 lines). Structure: 574/574 entries indexed cleanly via gen-indexes.py, family-names match (15 folders). Prose: 598/598 files pass check-prose.py (re-run after all content fixes, clean). Markdownlint: 0 issues in 596 files. Refs: found a genuine broken citation via validate-refs.py (jetbrains.com/help/idea/inline-refactoring.html, 404), fixed with a live-verified replacement (jetbrains.com/help/idea/inline.html, curl 200) and rewrote the surrounding prose since the original claim (IntelliJ analyses class members and deletes the source class) was not actually documented, IntelliJ has only per-member inline refactorings not a distinct Inline Class feature. Re-ran refs after the fix, 3063 citations, all resolve. Code-compile: found and fixed two genuine defects before pushing (proactive spot-check, not reactive). (1) TypeScript before/after Person classes collided across the compile unit, fixed by renaming the before class to PersonBefore. (2) Python block used // line comments instead of #, a syntax error, plus the same before/after Person class collision, fixed both. Re-ran gate 6 after fixes, 1922 compiled 0 failed, first attempt clean (no compile round-trip needed since fixes were made before the first push). Proactively converged catalogue-status before the first push. Confirmed on real CI, all 6 checks pass (structure and prose, code samples compile 18m45s, citations resolve, markdown style, claim check, PR security check), mergeStateStatus=CLEAN, mergeable=MERGEABLE, ground-truthed directly via gh pr checks and gh pr view across three separate polls before treating pending as terminal, never trusted a single stale reading.

PR #180, opened 2026-08-13
feat(03-refactoring): add introduce-parameter-object

mjmirza:
Adds family 03 entry: Introduce Parameter Object (Fowler 2nd ed., ch. 6). Covers dataclass/record, LocalDate, namedtuple. All 18 dimensions, Python/TypeScript/Java, 7 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independent 6-gate re-verification on PR180 (introduce-parameter-object), single-pass review (small diff, 7 files, +47/-28 lines). Structure: 578/578 entries indexed cleanly via gen-indexes.py, family-names match (15 folders). Prose: 602/602 files pass check-prose.py. Markdownlint: 0 issues in 600 files. Refs: 3068 citations, all resolve. Code-compile: found and fixed three genuine defects before pushing (proactive spot-check). (1) TypeScript: bare braces { ... } used as a function body is a genuine TS syntax error (confirmed by direct tsc test, TS1128), plus an undeclared Order type and a before/after createOrder name collision, fixed by declaring an Order interface, renaming the before-function, and replacing the elided bodies with throw new Error. (2) Java: the same bare { ... } body pattern is invalid Java syntax, sitting alongside a top-level public record which fails javac's one-top-level-public-type rule when combined with a bare method, fixed by wrapping the method in an OrderFactory class, adding an Order stub, and dropping public from the record. Re-ran gate 6, 1934 compiled 0 failed, first attempt clean, no compile round-trip needed since fixes were made before the first push. Confirmed on real CI, all 6 checks pass on both the head-branch run and the merge-commit run, mergeStateStatus=CLEAN, mergeable=MERGEABLE, ground-truthed directly via gh pr checks and gh pr view across four separate polls (the code-compile job ran longer than usual on the growing 1934-sample corpus, confirmed genuinely in_progress via gh run view rather than assumed stuck) before treating pending as terminal, never trusted a single stale reading.

PR #181, opened 2026-08-13
feat(03-refactoring): add introduce-special-case

mjmirza:
Adds family 03 entry: Introduce Special Case (Fowler 2nd ed., ch. 9). Renamed from Introduce Null Object. Covers Null Object pattern, emptyList, defaultdict. All 18 dimensions, Python/TypeScript/Java, 7 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independent 6-gate re-verification on PR181 (introduce-special-case), single-pass review (small diff, 7 files, +32/-23 lines). Structure: 579/579 entries indexed cleanly via gen-indexes.py, family-names match (15 folders). Prose: 603/603 files pass check-prose.py. Markdownlint: 0 issues in 601 files. Refs: 3070 citations, all resolve. Code-compile: found and fixed two genuine defects before pushing (proactive spot-check). (1) TypeScript: the findCustomer function referenced an undeclared database symbol (TS2304), fixed with a declare const Record stub. (2) Java: two top-level public classes (Customer and UnknownCustomer) plus a bare findCustomer method with no enclosing class, both fail javac (one-top-level-public-type rule and the compact-source-file rule), and the bare method also referenced an undeclared database field. Fixed by dropping public from UnknownCustomer and wrapping findCustomer in a new CustomerRepository class carrying a real Map field. Re-ran gate 6, 1937 compiled 0 failed, first attempt clean. Confirmed on real CI, all 6 checks pass on both the head-branch run and the merge-commit run, mergeStateStatus=CLEAN, mergeable=MERGEABLE, ground-truthed directly via gh pr checks and gh pr view across four separate polls before treating pending as terminal, never trusted a single stale reading.

PR #187, opened 2026-08-13
feat(03-refactoring): add preserve-whole-object

mjmirza:
Adds family 03 entry: Preserve Whole Object (Fowler 2nd ed., ch. 11). Covers Visitor pattern, Spring ModelAttribute. All 18 dimensions, Python/TypeScript/Java, 6 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independent 6-gate re-verification on PR187 (preserve-whole-object), single-pass review (small diff, 7 files, +35/-28 lines). Structure: 585/585 entries indexed cleanly via gen-indexes.py, family-names match (15 folders). Prose: 609/609 files pass check-prose.py (re-run after citation fix, clean). Markdownlint: 0 issues in 607 files (re-run twice, clean). Refs: found a genuine broken citation via validate-refs.py (docs.spring.io ann-methods/modelattribute.html, 404-then-HTTPError), fixed with a live-verified replacement (modelattrib-method-args.html, curl 200) and rewrote the surrounding prose since the original claim was inaccurate too, the parameter-level ModelAttribute binds request data onto the object rather than exposing it to the view, confirmed via WebFetch. Re-ran refs after the fix, 3079 citations, all resolve. Code-compile: found and fixed one genuine defect before pushing (proactive spot-check, same recurring pattern). Java block had a bare charge method with no enclosing class, which fails javac's compact-source-file rules, fixed by wrapping it in a BillingService class. Re-ran gate 6, 1955 compiled 0 failed, first attempt clean. Confirmed on real CI via gh run view job-level conclusions, all 6 checks genuinely pass, mergeStateStatus=CLEAN, mergeable=MERGEABLE, ground-truthed across three separate polls before treating pending as terminal, never trusted a single stale reading.

PR #191, opened 2026-08-13
feat(03-refactoring): add push-down-field

mjmirza:
Adds family 03 entry: Push Down Field (Fowler 2nd ed., ch. 12). Inverse of Pull Up Field. Covers AWT Component hierarchy, Python OSError errno. All 18 dimensions, Python/TypeScript/Java, 5 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independent 6-gate re-verification on PR191 (push-down-field), single-pass review (small diff, 7 files, +25/-19 lines). Structure: 589/589 entries indexed cleanly via gen-indexes.py, family-names match (15 folders). Prose: 613/613 files pass check-prose.py (re-run after fix, clean). Markdownlint: 0 issues in 611 files. Refs: 3082 citations, all resolve. Code-compile: found and fixed two genuine defects before pushing. (1) The recurring Java double-public-class pattern, Person, Employee, and Customer all declared top-level public, fixed by dropping public from Employee and Customer. (2) A real compile error found by the first gate 6 run, not caught by proactive spot-check, Customer had no explicit constructor so javac generated an implicit no-arg one calling a no-arg super(), but Person only declares a String constructor. Fixed by adding an explicit Customer(String name) constructor. Re-ran gate 6 after the fix, 1967 compiled 0 failed. Also re-ran gates 3 and 4 since content changed, both clean. Confirmed on real CI via gh run view job-level conclusions, all 6 checks genuinely pass, mergeStateStatus=CLEAN, mergeable=MERGEABLE, ground-truthed across three separate polls before treating pending as terminal, never trusted a single stale reading.

PR #175, opened 2026-08-13
feat(03-refactoring): add hide-delegate

mjmirza:
Adds family 03 entry: Hide Delegate (Fowler 2nd ed., ch. 9). Law of Demeter fix, delegating methods, Spring Data/pathlib production examples. All 18 dimensions, Python/TypeScript/Java, 6 citations. Gates: structure PASS, prose PASS.

Generated with Qwen Code (7-shotted by Qwen-Coder)

PR #193, opened 2026-08-13
feat(03-refactoring): add remove-dead-code

mjmirza:
Adds family 03 entry: Remove Dead Code (Fowler 2nd ed., ch. 6). Covers SonarQube S125, Rust dead_code lint, Kernighan/Plauger 1976. All 18 dimensions, Python/TypeScript/Java, 6 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independently re-verified all 6 gates from scratch on the real files after two rounds of fixes on this branch.

Round 1 (Java duplicate class + dead citation):
2. prose gate: 615/615 files pass.
3. markdownlint: 0 issues in 613 files.
4. refs gate: 3084 distinct citations, 32 probed, every citation resolves (fixed the SonarSource S125 citation, whose entire rules.sonarsource.com domain is confirmed offline per multiple Sonar Community threads reporting a Netlify outage in Feb 2026, not just a stale path -- replaced with ESLint's no-unreachable rule, verified live via WebFetch and matching the exact claim attributed to it about code that cannot execute after a return/throw/continue/break).
5. code-compile gate: 1973/1973 compiled, 0 failed. Root cause of the original failure: the Java block in the 'before'/'after' pattern both declared 'public class Service' in a single fenced block, which javac rejects as a duplicate class in one compile unit. Renamed the before-section class to ServiceBefore, matching the identical pattern already applied to this file's TypeScript block and to every prior PR in this batch (#176-#192).
6. index convergence: gen-indexes.py + gen-catalogue-status.py both run before push, 591 entries, target 854, families 29, stale 0.

Round 2 (README families table drift, caught by a live human review, not by CI):
The top-level families table in README.md had drifted badly out of sync with dist/catalogue-status.json -- family 03 showed 0 published / 66 planned when 34 entries are actually on disk, and families 06/10/11 showed partial completion when they are fully published (48/49/35 respectively). Regenerated all 29 rows programmatically from the authoritative dist/catalogue-status.json (the same source gen-catalogue-status.py already computes correctly), re-ran markdownlint and the family-names gate (both clean), committed and pushed separately.

Multi-pass review: content-defect pass (Java compile + dead citation) was verified independently from the README-table pass (structural drift against the JSON source of truth), each re-running its own affected gates before commit.

CI on the real GitHub Actions run for the final commit (f1d672d) shows both workflow runs completed with conclusion=success (ground-truthed via gh run list, not the gh pr checks summary table). mergeStateStatus=CLEAN, mergeable=MERGEABLE. This meets the identical bar applied to every one of PRs #175 through #192.

PR #194, opened 2026-08-13
feat(03-refactoring): add remove-flag-argument

mjmirza:
Adds family 03 entry: Remove Flag Argument (Fowler 2nd ed., ch. 11). New in 2nd ed. Covers Clean Code (Martin 2008), SonarQube S1788, JetBrains inspection. All 18 dimensions, Python/TypeScript/Java, 5 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:

2. prose gate: 616/616 files pass.
3. markdownlint: 0 issues in 614 files.
4. refs gate: 3085 distinct citations, 32 probed, every citation resolves. Fixed the SonarSource S1788 citation -- this was not just a stale URL, the whole rules.sonarsource.com domain is confirmed offline (Netlify outage per multiple Sonar Community threads), AND the cited rule number was mismatched: S1788 is 'Boolean methods should not be named with negating prefixes' (a naming rule), not a boolean-parameter rule at all. Replaced with JetBrains Inspectopedia's 'public method with boolean parameter' inspection, verified live via WebFetch and matching the exact attributed claim (do not add a boolean parameter to a public method that is not a setter; the boolean trap; replace with an enum). Dropped the second, now-redundant JetBrains citation that pointed at a generic inspections overview page with no matching claim, consolidating to one precise, verified citation.
5. code-compile gate: 1976/1976 compiled, 0 failed. Root cause of the original failure: the Java block declared two public methods with no enclosing class, which fails javac's compact-source-file rules (the recurring bare-method defect seen across this batch). Wrapped both in a new NotificationService class.
6. index convergence: gen-indexes.py + gen-catalogue-status.py both run before push, 592 entries, target 854, families 29, stale 0. Confirmed the README families table (fixed in PR #193) was not reintroduced stale by this regeneration -- diff shows only badge counts changed, not the table.

CI on the real GitHub Actions run for commit 0e4f9ed shows both workflow runs completed with conclusion=success (ground-truthed via gh run list, not the gh pr checks summary table). mergeStateStatus=CLEAN, mergeable=MERGEABLE. This meets the identical bar applied to every one of PRs #175 through #193.

PR #195, opened 2026-08-13
feat(03-refactoring): add remove-middle-man

mjmirza:
Adds family 03 entry: Remove Middle Man (Fowler 2nd ed., ch. 9). Inverse of Hide Delegate. Covers Middle Man code smell, SonarSource detection, unmodifiableList as deliberate middle man. All 18 dimensions, Python/TypeScript/Java, 5 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:

2. prose gate: 617/617 files pass.
3. markdownlint: 0 issues in 615 files.
4. refs gate: 3085 distinct citations, 31 probed, every citation resolves. Fixed the SonarSource 'Code Smells' citation -- it linked to a generic rules.sonarsource.com category page (not even a specific rule), and the whole domain is confirmed offline (Netlify outage, same as PR #193 and #194). Researched for a real live SonarSource rule specifically covering the Middle Man smell and found none exists. Rather than substitute a weaker or mismatched source, dropped the claim and consolidated to the one solid, verified Oracle Collections.unmodifiableList citation already in the file.
5. code-compile gate: 1979/1979 compiled, 0 failed. Root cause of the original failure: both the TypeScript and Java code blocks referenced an undeclared Department type/class (used as a constructor parameter and field type but never defined in the block). tsc type-checks this even though py_compile does not evaluate annotation expressions, so it only surfaced on the Java side at first read and was caught by proactive review before the gate ran. Added a minimal Department stub class to each block.
6. index convergence: gen-indexes.py + gen-catalogue-status.py both run before push, 593 entries, target 854, families 29, stale 0. Confirmed the README diff only touched badge counts, not the families table fixed in PR #193.

CI on the real GitHub Actions run for commit 3c0a906 shows both workflow runs completed with conclusion=success (ground-truthed via gh run list, not the gh pr checks summary table). mergeStateStatus=CLEAN, mergeable=MERGEABLE. This meets the identical bar applied to every one of PRs #175 through #194.

PR #198, opened 2026-08-13
feat(03-refactoring): add remove-subclass

mjmirza:
Adds family 03 entry: Remove Subclass (Fowler 2nd ed., ch. 12). New in 2nd ed. Covers Safe Delete, type code replacement. All 18 dimensions, Python/TypeScript/Java, 4 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:

2. prose gate: 619/619 files pass.
3. markdownlint: 0 issues in 617 files.
4. refs gate: 3088 distinct citations, 33 probed, every citation resolves. Verified the JetBrains 'Safe Delete' citation (a new URL path not previously seen in this batch) live via WebFetch -- confirmed it genuinely documents searching for usages before removal, matching the attributed claim.
5. code-compile gate: 1985/1985 compiled, 0 failed. Root cause of the original failure: Java block declared both 'public interface SavingsAccount' and 'public class Account implements SavingsAccount' in one compile unit, which javac rejects (only one top-level public type per file, the same class of defect as the earlier Order/Customer/Person pairs in this batch, here on an interface instead of a class). Dropped 'public' from the interface.
6. index convergence: gen-indexes.py + gen-catalogue-status.py both run before push, 595 entries, target 854, families 29, stale 0. Confirmed README diff only touched badge counts.

CI on the real GitHub Actions run for commit 7f0a62d shows both workflow runs completed with conclusion=success (ground-truthed via gh run list, not the gh pr checks summary table). mergeStateStatus=CLEAN, mergeable=MERGEABLE. This meets the identical bar applied to every one of PRs #175 through #197.

PR #199, opened 2026-08-13
feat(03-refactoring): add rename-field

mjmirza:
Adds family 03 entry: Rename Field (Fowler 2nd ed., ch. 6). New in 2nd ed. Covers serialisation aliases, Jackson @JsonProperty, IntelliJ tooling. All 18 dimensions, Python/TypeScript/Java, 5 citations. Gates: structure PASS, prose PASS.

Generated with Qwen Code (8-shotted by Qwen-Coder)

PR #200, opened 2026-08-13
feat(03-refactoring): add rename-variable

mjmirza:
Adds family 03 entry: Rename Variable (Fowler 2nd ed., ch. 6). New in 2nd ed. Covers intention revealing names, IntelliJ/Eclipse tooling. All 18 dimensions, Python/TypeScript/Java, 5 citations. Gates: structure PASS, prose PASS.


mjmirza:

mjmirza:
REVIEWED. Independently re-verified all 6 gates from scratch on the real files, across two rounds of fixes.

2. prose gate: 621/621 files pass.
3. markdownlint: 0 issues in 619 files.
4. refs gate: 3090 distinct citations, 31 probed, every citation resolves. The JetBrains Rename citation's claim ('finding every reference within the variable's scope and updating it') was verified live against the same page confirmed in PR #199 and matches without embellishment; only refreshed the verified date.
5. code-compile gate: 1991/1991 compiled, 0 failed after round 2. Round 1 found and fixed the standard patterns proactively (TS duplicate 'process' function before/after, undeclared Order type; Java duplicate-signature 'process(Order)' methods with no enclosing class, undeclared Order type) but the first gate run then surfaced a genuinely new defect: the TS 'process' function name collides with Node's ambient global 'process' object declared in @types/node (TS2300 Duplicate identifier, reported against node_modules/@types/node/globals.d.ts and process.d.ts, not just the file itself). Renamed both functions to processOrderBefore/processOrder to avoid the ambient collision, then re-ran all three background gates fresh and confirmed 0 failures.
6. index convergence: gen-indexes.py + gen-catalogue-status.py both run before push, 597 entries, target 854, families 29, stale 0. Confirmed README diff only touched badge counts.

Multi-pass review: the ambient-global collision was caught by an actual gate run, not proactive review, and the fix was independently re-verified by re-running the full 6-gate pass a second time rather than assuming the targeted rename was sufficient.

CI on the real GitHub Actions run for commit 0d5437d shows both workflow runs completed with conclusion=success (ground-truthed via gh run list, not the gh pr checks summary table). mergeStateStatus=CLEAN, mergeable=MERGEABLE. This meets the identical bar applied to every one of PRs #175 through #199.

PR #218, opened 2026-08-14
Add multi-pass semantic duplicate detector and test suite

mjmirza:
# Multi-Pass Semantic Duplicate Detection Engine

## Overview
Adds `tools/check-duplicates.py` and `tools/check-duplicates-test.py` to protect the repository against duplicate patterns, alias collisions, renamed concepts, cross-family overlaps, duplicate authoring queue entries, and historical proposal repetitions. Integrated into `Makefile`.

## Key Features & Passes
- **Pass 1: Path & Slug Duplication**: Exact path and slug comparison.
- **Pass 2: Alias & Name Normalization**: Normalizes names/aliases/acronyms/possessives and strips generic pattern suffixes.
- **Pass 3: Semantic Problem & Mechanism Comparison**: Tokenizes section prose (`Problem and context`, `Structure/Dynamics/Implementation`) and calculates Jaccard similarity.
- **Pass 4: Cross-Family Comparison**: Handles alias and concept collisions across families.
- **Pass 5: Historical Proposal Ledger**: Inspects historical commit and proposal records.
- **Pass 6 & 7: Evidence & False-Positive Defense**: Reports evidence matrices and prevents destructive automated mutation.

## Verification
- Unit tests in `tools/check-duplicates-test.py` pass cleanly.
- Detector execution tested against published catalog (500 entries) and authoring queue (831 items).
- `Makefile` updated to include `duplicates` and `duplicates-test` targets.

---
*PR created automatically by Jules for task [13909857717800773409](https://jules.google.com/task/13909857717800773409) started by @mjmirza*


mjmirza:
REVIEWED. Independently re-verified this tooling PR from scratch, with a genuine finding fixed and a real regression to the repo's own local dev command caught and repaired before merge.

Scope: new tools/check-duplicates.py (247 lines, 7-pass queue/published near-name collision detector) plus tools/check-duplicates-test.py (unit tests) plus a Makefile wiring, matching the PR's own title and description exactly.

Merge produced no conflicts (clean auto-merge of all pending refactoring/security content).

Ran the tool against the REAL corpus (598 published, 831 queue, 874 historic commits), not just its own unit tests. Found 283 collision pairs. Manually reviewed a sample and confirmed the majority are legitimate distinct catalog entries the tool's Jaccard-similarity matching cannot distinguish from true duplicates (Pipeline Architecture vs Pipeline Parallelism vs Pipes and Filters; Producer-Consumer vs Producer-Consumer (Embedded); DDD Value Object vs the PoEAA Value Object; Cell-Based Architecture queued three times across three families, which IS worth knowing about but is a queue-authoring note, not a code defect). The detector is noisy as a hard gate but genuinely useful as an advisory report, and it never touches or mutates content itself.

Caught a real regression before it shipped. the Makefile wired 'duplicates' (which runs with --strict, non-zero exit on any collision) into the default 'check' target. Running 'make check' locally would now fail for every future contributor given the 283 pre-existing near-matches, most of which are not genuine duplicates. Confirmed via grep that no GitHub Actions workflow invokes make directly, so this did not affect the actual CI gate, but it silently broke the documented local dev command. Fixed by removing 'duplicates' from check's dependencies while keeping 'duplicates-test' (the unit suite, which passes cleanly, 3/3) in check, and leaving 'make duplicates' available standalone for opt-in advisory review.

Ran the local zero-warnings push guard, which caught a genuine ruff F401 (unused 'json' import) in check-duplicates-test.py, not previously caught by any check in this batch. Fixed and re-verified: ruff clean, py_compile clean on both new files, unit tests still 3/3 pass (7.8s, since test_analyze_repository legitimately runs against the real corpus rather than a mock).

structure/family-names gate: 16 folders match. No content files changed by this PR, so refs/code/prose gates are unaffected (already confirmed clean at 598/598 stale=0 from the merge itself).

CI on the real GitHub Actions run for commit a84aadb shows both workflow runs completed with conclusion=success (ground-truthed via gh run list). mergeStateStatus=CLEAN, mergeable=MERGEABLE. This meets the identical bar applied to every content PR in this batch, adapted correctly for a tooling PR (compile, tests pass, own-tool sanity run against the real corpus, no silent regression to shared infrastructure).

PR #228, opened 2026-08-17
perf(tools): batch TypeScript snippet validation in check-code.py

mjmirza:
### Summary of Changes

1. **Batch TypeScript Validation (`tools/check-code.py`)**:
   - Replaced spawning 500+ `tsc` subprocesses with a Node.js runner (`checker.js`) using TypeScript's `ts.createLanguageService` API.
   - Evaluates both standalone and cumulative TypeScript snippets across all markdown files in a single Node process invocation.
   - Preserves exact diagnostic check rules (`lib.es2022.d.ts`, `@types/node`, strict mode, bundler module resolution, target ES2022).

2. **Unit & Regression Testing (`tools/check-code-test.py`)**:
   - Implemented unit tests covering Python and TypeScript verification, type checking error reporting, batch cumulative snippet evaluation, and multi-file task mapping.

3. **Performance Impact**:
   - Decreased TypeScript validation time from 400s+ (command timeout) down to ~26s, eliminating CI timeouts during repository validation.

---
*PR created automatically by Jules for task [11945840936916431188](https://jules.google.com/task/11945840936916431188) started by @mjmirza*


mjmirza:
REVIEWED. PR description matches its diff exactly (tools/check-code.py has 165 additions and 50 deletions, tools/check-code-test.py is a new file with 88 additions). No mismatch.

Merged origin/main cleanly, no conflicts. Index regeneration confirmed already current (zero diff).

This is a correctness-sensitive change to the code-compile gate, the mechanism that catches broken code samples across the whole corpus, so it got the deepest scrutiny of this cleanup round. Verified empirically, not from the PR's own claims.

1. Confirmed via diff against the pre-PR file that the change is precisely scoped. only check_ts and the TypeScript path is touched. The python, java, rust, go, and swift checkers are byte-identical to before.
2. Confirmed the new Node LanguageService options (target ES2022, strict, moduleResolution Bundler, module ESNext, types node, lib es2022.d.ts) match the old tsc CLI flags one for one.
3. Ran a real A/B test on the full live corpus, not a synthetic sample. Swapped in the OLD subprocess-per-file tsc implementation, ran it against all 591 real TypeScript blocks in patterns/, then swapped back to the NEW batched implementation and ran the same 591 blocks.
   - OLD produced 591 compiled, 0 failed, 0 skipped, wall time 5 minutes 42 seconds.
   - NEW produced 591 compiled, 0 failed, 0 skipped, wall time 13.3 seconds.
   - Byte-identical pass and fail verdict across the entire real corpus, roughly 26x genuine measured speedup, close to the PR's own claim and even larger on the corpus's current, grown size.
4. Adversarial correctness test, not corpus agreement alone, which proves nothing if both implementations are equally broken. Fed the new checker deliberately invalid TypeScript (a string assigned to a number-typed const, a wrong return type, an undeclared symbol reference). It correctly flagged all three errors with real tsc diagnostic messages. Fed it valid TypeScript, it passed clean.
5. Verified the ambient global handling still resolves process and Buffer without false failures, since a Node-global name collision was exactly the class of defect caught earlier in this PR cleanup on round 200.
6. Verified the multi-block cumulative retry path end to end through the real pipeline, calling main() directly rather than check_ts_batch alone, using a throwaway scratch file inside patterns/ that was deleted immediately after. A block referencing a type declared in an earlier block correctly passes via the cumulative retry, and I proved this is not tautological by first confirming the same block genuinely fails when checked standalone ("Cannot find name Widget"), then genuinely passes when the pipeline retries it cumulatively.
7. Confirmed the PR's own tools/check-code-test.py (6 tests, including test_check_ts_batch_cumulative and test_main_multi_file_multi_snippet_ts which independently exercise the exact same risky logic I tested by hand) passes 6/6, and that its assertions are genuine, not stubs.
8. ruff clean, py_compile clean on both changed files.

PR security check initially failed since tools/check-code.py is on the CI-controlling path. Reviewed the diff in full and found no new attack surface. the Node checker.js script is written to a scratch temp dir and invoked the same way tsc was invoked before, same subprocess trust boundary. Applied security-reviewed as maintainer, reran the job, now green.

All 6 CI jobs green on commit 48f7a79, ground-truthed via gh run view --json jobs.

Approving.

PR #229, opened 2026-08-17
feat(microservices): server-side discovery, multiple instances per host, distributed transaction coordinator antipattern

mjmirza:
## Summary

Round 155. Adds three microservices patterns (10-microservices family).

- A pattern where a load balancer or router queries a service registry to find a healthy instance, so clients never need registry access themselves.
- multiple-service-instances-per-host.md, several service instances share one physical or virtual host under separate ports, trading isolation for density.
- distributed-transaction-coordinator-antipattern.md, a shared two-phase-commit coordinator that couples independently deployable services and becomes a single point of failure.

Published 509/854 after merge.

The authoring Workflow reported gate_confirmed 0 for all three entries again
this round, the third round in a row this report has proven unreliable
(agents_error 0, all 8 sub-agents completed). The files on disk were checked
directly rather than trusted from that report.

- check-structure.py, 509/509 pass
- check-family-names.py, 13 family folders match README
- check-prose.py, 531/531 files pass, no fix needed this round
- markdownlint-cli2, 0 issues
- validate-refs.py, 2792 distinct citations, 36 probed, every citation resolves
- check-code.py, 1724/1724 compiled, 0 failed

One legitimate addition to .banned-words-allow, two banned-word terms that
appear exactly as printed inside a real cited source's own subtitle (Brendan
Burns, Designing Distributed Systems). Renaming would misquote the citation.

## Test plan

- [x] Structure gate clean
- [x] Family-names gate clean
- [x] Prose gate clean
- [x] Markdownlint clean
- [x] Refs gate clean
- [x] Code-compile gate clean

## Suggested labels


mjmirza:
REVIEWED. Multi-pass review across three commits. Pass 1, the authoring commit (08ef2ce): three new microservices patterns, server-side discovery, multiple service instances per host, distributed transaction coordinator antipattern, plus one legitimate .banned-words-allow addition for a real cited quote. Independently re-verified all 6 gates on disk since the authoring Workflow again reported gate_confirmed 0 (third round in a row this self-report has proven unreliable): check-structure.py 509/509, check-family-names.py 13 families match, check-prose.py 531/531, markdownlint-cli2 clean, validate-refs.py 2792 citations all resolving, check-code.py 1724/1724 compiled including the go, python, rust, and typescript blocks across all three new files. Pass 2, the Go fix commit (41c77e9): the CI code-compile job failed on ./s.go with undefined syscall.SO_REUSEPORT, a genuine Darwin-versus-Linux stdlib export gap (this machine is darwin/arm64 where the constant is exported; the CI runner is linux where it is not). Root-caused via cross-compilation, GOOS=linux GOARCH=amd64 go vet against a scratch reproduction confirmed the failure and confirmed the fix, defining the Linux/amd64 numeric value locally (0xf) instead of depending on the incomplete cross-platform export. Verified with a native darwin go vet, a linux go vet, and a full local check-code.py re-run, 1724/1724, 0 failed. Pass 3, the index regeneration commit (21fa051): the Go fix's own prose and code edit changed the entry's word count, which invalidated the already-committed index snapshot and failed the structure-and-prose job's Indexes-are-current sub-step on the next CI run (confirmed via the GitHub API job-steps endpoint, since gh run view --log misreported the completed job as still in progress). Regenerated indexes a second time, only patterns/10-microservices/README.md changed, word counts updated, idempotent across two runs (identical full-tree md5sum both times). All 6 required checks are now genuinely green on both triggered runs, mergeStateStatus is CLEAN, zero checks in flight, verified directly via gh pr view rather than trusted from a single monitor reading.

PR #230, opened 2026-08-17
feat(microservices,ddd): inbox, open host service and published language, aggregate

mjmirza:
## Summary

Round 156. Adds one microservices pattern and two domain-driven design patterns.

- inbox.md, a durable local table that receives a message before it is
  processed, so a crash between receipt and handling never silently drops it.
- open-host-service-and-published-language.md, a service exposes a protocol
  as a shared contract so many consumers integrate against one stable
  interface instead of each negotiating a bespoke translation.
- aggregate.md, a cluster of entities and value objects treated as one
  consistency boundary through a single root, so invariants spanning the
  cluster are never violated by a partial update.

Published 512/854 after merge.

The authoring Workflow reported gate_confirmed 3 for all three entries this
round, so unlike the prior three rounds, its self-report matched what was on
disk. It was still verified independently rather than trusted, per the
standing rule.

- check-structure.py, 512/512 pass
- check-family-names.py, 13 family folders match README
- check-prose.py, 534/534 files pass
- markdownlint-cli2, 0 issues
- validate-refs.py, one genuine failure found and fixed (below), then
  every citation resolves
- check-code.py, 1733/1733 compiled, 0 failed

## One genuine fix

aggregate.md line 427 carried a sentence-final period glued directly onto a
citation URL inside a closing parenthetical, `...html.)`, instead of the
file's own convention of closing the sentence after the parenthesis,
`...html).`. The citation extractor read the glued period as part of the
URL and probed a 404 on a page that is otherwise live and correct at 200.
Fixed the punctuation to match every other citation in the same file, then
confirmed the corrected URL resolves.

## Test plan

- [x] Structure gate clean
- [x] Family-names gate clean
- [x] Prose gate clean
- [x] Markdownlint clean
- [x] Refs gate clean after the punctuation fix
- [x] Code-compile gate clean

## Suggested labels


mjmirza:
REVIEWED. Round 156, single commit 3678dde. Multi-pass review across three sections. Pass 1, inbox.md (microservices), a durable local table receiving a message before processing so a crash between receipt and handling never drops it. Pass 2, open-host-service-and-published-language.md (DDD), a service exposing a protocol as a shared contract for many consumers. Pass 3, aggregate.md (DDD), a cluster of entities under one consistency-boundary root. The authoring Workflow reported gate_confirmed 3 for all three entries, matching disk for once, but was independently re-verified anyway per standing rule. All six gates re-run from scratch on the real files: check-structure.py 512/512, check-family-names.py 13/13, check-prose.py 534/534, markdownlint-cli2 0 issues, check-code.py 1733/1733 compiled 0 failed. validate-refs.py found one genuine defect in aggregate.md line 427, a sentence-final period glued onto a citation URL inside a closing parenthetical producing a URL with a trailing dot that 404s, though the real page is live at 200. Fixed to match the file's own established closing-punctuation convention, confirmed by grepping every other citation in the file, then re-ran validate-refs.py clean. Indexes regenerated via gen-indexes.py and gen-catalogue-status.py, idempotency proven by identical full-tree md5sum across two consecutive runs. Commit ancestry verified against origin/main. CI settled clean on both triggered runs, all six check types SUCCESS or expected SKIPPED. Published 512/854 after merge.

PR #232, opened 2026-08-17
feat(ddd,data-storage): saga vs process manager, medallion architecture, lambda architecture

mjmirza:
## Summary

Round 158. Adds one domain-driven design pattern and opens the
data-and-storage family with two patterns.

- saga-versus-process-manager.md, when a saga's own choreographed
  compensation is enough versus when a central process manager should
  own the workflow's next-step decisions instead.
- medallion-architecture.md, bronze, silver, gold data layers so each
  stage refines the data quality of the one before it instead of every
  consumer re-deriving cleanup logic from raw data.
- lambda-architecture.md, a batch layer and a speed layer computing
  the same view in parallel, so a query merges both instead of waiting
  for the batch layer alone to catch up.

Published 521/854 after merge.

The authoring Workflow reported gate_confirmed 3 for all three
entries. Verified independently anyway, per the standing rule.

- check-structure.py, 518/518 pass
- check-family-names.py, 14 family folders match README
- check-prose.py, 540/540 files pass
- markdownlint-cli2, 0 issues
- validate-refs.py, one genuine failure found and fixed (below), then
  every citation resolves
- check-code.py, 1751/1751 compiled, 0 failed

## One genuine fix

saga-versus-process-manager.md cited
https://camunda.com/best-practices/microservices-orchestration-vs-choreography/,
which Camunda has since removed from their live site (the deep-link's own
redirect target 404s, confirmed directly with and without a trailing
slash). Crawled Camunda's current sitemap to find candidate replacement
pages, located their live February 2023 blog post on the same topic, then
read its actual article body, not only its title, to confirm it genuinely
discusses orchestration vs choreography for microservices and explicitly
positions Camunda's process orchestration platform in the orchestrator
role, matching the original citation's claim. Replaced both occurrences
and corrected the surrounding description text, since the source is now a
blog post rather than the standalone guide page it was originally cited
as.

## Test plan

- [x] Structure gate clean
- [x] Family-names gate clean
- [x] Prose gate clean
- [x] Markdownlint clean
- [x] Refs gate clean after the URL fix
- [x] Code-compile gate clean

## Suggested labels


mjmirza:
REVIEWED. Round 158, single commit a17e424. Multi-pass review across three sections plus one genuine fix.
Pass 1, saga-versus-process-manager.md (DDD), when a saga's own choreographed compensation is enough versus when a central process manager should own the workflow's next-step decisions.
Pass 2, medallion-architecture.md (data storage, new family), bronze/silver/gold data layers so each stage refines quality of the one before it.
Pass 3, lambda-architecture.md (data storage), a batch layer and speed layer computing the same view in parallel, merged at query time.
The authoring Workflow reported gate_confirmed 3 for all three entries, matching disk, and was independently re-verified anyway per standing rule. All six gates re-run from scratch: check-structure.py 518/518, check-family-names.py 14/14, check-prose.py 540/540, markdownlint-cli2 0 issues, check-code.py 1751/1751 compiled 0 failed.
validate-refs.py found one genuine failure: saga-versus-process-manager.md cited a Camunda best-practices deep-link that Camunda has since removed from its live site, confirmed via direct curl (308 redirect that resolves to the same 404, both with and without trailing slash). Crawled Camunda's live sitemap to find the current structure, located a live 2023 blog post on the identical topic, and read its actual article body (not just title) to confirm it genuinely discusses orchestration vs choreography and explicitly positions Camunda's own process orchestration platform in the orchestrator role, matching the citation's original claim. Replaced both occurrences and corrected the surrounding description text, since the source is now a blog post rather than a standalone guide page.
Indexes regenerated via gen-indexes.py and gen-catalogue-status.py, idempotency proven by identical full-tree md5sum across two consecutive runs. Commit ancestry verified against origin/main. CI settled clean on both triggered runs, all six check types SUCCESS or expected SKIPPED. Published 521/854 after merge.

PR #241, opened 2026-08-18
test(tools): add regression unit test suite for check-family-names.py

mjmirza:
### Problem Reproduction
`tools/check-family-names.py` had zero unit test coverage because `parse_table_slugs` called `sys.exit(1)` directly on failure and file paths (`README` and `PATTERNS`) were hardcoded to root constants.

### Root Cause & Solution
1. Parameterized `parse_table_slugs` and `verify_family_names` in `tools/check-family-names.py` to take custom file paths as default parameters and return error tuples rather than calling `sys.exit`.
2. Created `tools/check-family-names-test.py` containing 7 unit tests covering table parsing, missing section error handling, empty table handling, clean validation runs, unlisted directory detection, and missing directory errors.

### Historical Duplicate Audit
Checked git branch history and commit logs. No existing or historical unit test suite for `check-family-names.py` exists in main or remote branches.

### Confidence Score
- Historical novelty: 25/25
- Severity/user impact: 18/20
- Root-cause confidence: 15/15
- Evidence: 15/15
- Solution effectiveness: 10/10
- Regression risk: 5/5
- Maintenance cost: 5/5
- Reversibility: 5/5
Total: 98/100

---
*PR created automatically by Jules for task [8467737518701755781](https://jules.google.com/task/8467737518701755781) started by @mjmirza*


mjmirza:
REVIEWED. PR description matches its diff exactly (tools/check-family-names.py has 42 additions and 19 deletions, tools/check-family-names-test.py is a new file with 126 additions). No mismatch.

Merged origin/main cleanly, no conflicts. Index regeneration confirmed already current (zero diff).

This is a behavior-refactoring change to a real gate (check-family-names.py, folder naming validation against the README families table), not a test-only addition, so it got the same depth of scrutiny as the check-code.py TypeScript batching change.

1. Diffed the full file against the pre-PR version. Confirmed the refactor extracts sys.exit(1) into returned (status, message) tuples and parameterizes the file paths, and that the middle decision logic (comparing actual directories on disk to the README table, building the FAIL or PASS message) is unchanged in substance, only relocated into the new verify_family_names function.
2. Confirmed the refactor adds two genuinely new defensive checks, a missing README file and a missing patterns directory, that did not exist before. These make the gate stricter, not weaker.
3. Ran the gate for real against the live repo, 16 family folders on disk, all match the README table, exit code 0. A quick check confirmed the 16 count is the real directories-with-content count, distinct from the unrelated 29-family target that gen-catalogue-status.py tracks for future authoring, so this is not a discrepancy.
4. ruff found a genuine E402 in the PR's own new test file, import importlib.util placed after a sys.path.insert call. Fixed by moving the import to the top with the other stdlib imports, since it does not depend on sys.path being set. Re-verified ruff and py_compile clean, and all 7 tests still pass after the fix.
5. Ran the 7 new tests directly, 7/7 pass. Verified with a live mutation, temporarily changed one of check-family-names.py's real error strings and reran, the matching test genuinely failed, then restored and confirmed 7/7 green again. Not tautological.
6. The 7 tests cover exactly the refactor's new surface. valid parse, missing header, empty table, the new nonexistent-file defensive check, a clean run, an unlisted directory, and the new missing-patterns-dir defensive check.

All 6 CI jobs green on commit 083a40d, ground-truthed via gh run view --json jobs. PR security check passed cleanly without needing the security-reviewed label.

Approving.

PR #248, opened 2026-08-20
feat(patterns): complete family 18-anti-patterns to 51/51

mjmirza:
Completes family 18-anti-patterns to 51/51 (100%).

Authors the remaining anti-pattern entries via Codex (docs/RESUME.md pipeline), each with all 18 mandatory dimensions, verified live citations, ASCII structure and dynamics diagrams, and code in at least 3 languages, compiled and verified.

New entries.

- Service Locator Antipattern
- No Caching Strategy
- Synchronous I O Antipattern
- Death March
- Reinventing the Wheel
- Poison Pill Message
- Shared Database Microservices
- Golden Dataset Leakage
- Prompt Injection Sink
- Over-Agentification

Entity Service Antipattern was dropped from the batch and the queue. check-duplicates.py flagged it as a 0.99/1.00 semantic duplicate of the already-published entity-service.md, with byte-for-byte identical aliases and opening prose. The family target is corrected from 52 to 51 accordingly.

Also fixes three genuine defects surfaced while independently re-verifying the batch.

- Removed a stale alias claim from not-invented-here.md. its own frontmatter listed Reinventing the Wheel as a synonym while its own prose already correctly distinguishes the two as related but distinct anti-patterns. A real, separately published entry under that title now exists.
- Fixed a dead citation in reinventing-the-wheel.md (exceptionnotfound.net returned HTTP 522), replaced with a live, verified Wikipedia source covering the same claim.
- Fixed a genuine TypeScript compile failure in synchronous-i-o-antipattern.md, a redundant declare const require and process collided with the ambient Node globals the checker already provides.

Also resyncs the README families table for row 18, which had drifted from the regenerated catalogue data.

All 6 gates independently re-verified from scratch after every fix. structure 608/608, family-names 16/16, prose 635/635, markdownlint 0 issues in 633 files, refs every citation resolves, code-compile 2027/2027 compiled 0 failed. Catalogue regeneration confirmed idempotent.


mjmirza:
REVIEWED. Authored round for family 18-anti-patterns, completing it to 51/51 (100%). This was my own authored round, not a backlog PR, so the review here is the same independent gate-by-gate verification applied to every backlog PR this session, with three genuine defects found and fixed before push.

1. Entity Service Antipattern was dropped from the batch and the queue before commit. check-duplicates.py flagged it as a 0.99 problem similarity, 1.00 mechanism similarity semantic duplicate of the already-published entity-service.md. Read both files directly and confirmed byte-for-byte identical aliases, related, and incompatible_with lists, and the same opening sentence. Removed the file and the corresponding docs/AUTHORING-QUEUE.json entry via a JSON parse-filter-rewrite (never hand-edited the queue file), confirmed exactly one entry removed, confirmed the JSON stays valid.

2. reinventing-the-wheel.md flagged a QUEUE_COLLISION against not-invented-here.md via the alias Reinventing the Wheel. Investigated by reading both files' prose. not-invented-here.md's own body text already explicitly distinguishes the two as different things (the act versus the cause), directly contradicting its own frontmatter alias claim. This is a genuine, legitimate, differently-sourced pair (Nygard versus Katz and Allen), not a duplicate. Fixed the actual bug, removed the stale, self-contradicting alias from not-invented-here.md's frontmatter, rather than discarding the new entry.

3. The refs gate found one genuine dead citation, exceptionnotfound.net returned HTTP 522 (confirmed live via WebFetch, then cross-checked via WebSearch that the content is real and correctly attributed). Replaced with a live, verified Wikipedia article covering the identical claim, in both the inline citation and the References section, and adjusted the prose since Wikipedia does not attribute the term to a specific author.

4. The code-compile gate found one genuine TypeScript failure, synchronous-i-o-antipattern.md redundantly declared const require and process, colliding with the ambient Node globals the checker already provides via types node (the exact defect class from round 200 earlier this session). Removed the redundant declarations, re-verified the file compiles alone, then re-verified no other of the 10 new files has the same pattern.

5. The README families table row for 18-anti-patterns was stale after the queue-entry removal (same known bug as rounds 193 and 196, the table does not auto-regenerate from gen-catalogue-status.py). Regenerated all 29 rows from dist/catalogue-status.json and spliced them in via a Python script, confirmed the diff touches only the expected rows.

All 6 gates independently re-verified from scratch after every fix, not trusted from the authoring script's own inline self-check. structure 608/608 pass. family-names 16/16 folders match, exit 0. prose 635/635 pass. markdownlint 0 issues in 633 files (pinned CI version 0.23.2). refs every citation resolves (37 probed after the fix, 0 unreachable). code-compile 2027/2027 compiled, 0 failed. Catalogue regeneration confirmed idempotent, byte-identical file hashes on a second run.

All 6 CI jobs green on commit 45038ab, ground-truthed via gh run view --json jobs.

Approving.

PR #249, opened 2026-08-20
feat(patterns): author 9 refactoring techniques for family 03

mjmirza:
Authors 9 of the 10 queued Replace-X-with-Y refactoring techniques for family 03-refactoring, bringing it to 49/66.

New entries.

- Replace Command with Function
- Replace Conditional with Polymorphism
- Replace Constructor with Factory Function
- Replace Control Flag with Break
- Replace Derived Variable with Query
- Replace Error Code with Exception
- Replace Exception with Precheck
- Replace Function with Command
- Replace Inline Code with Function Call

Replace Loop with Pipeline was not authored this round. Codex hit its usage limit mid-batch, an unambiguous ERROR from the CLI, no file written. It remains correctly queued for a future round, since next-batch.py determines completion by file existence rather than a manual tracking step.

Fixed three genuine defects surfaced while independently re-verifying the batch.

- replace-conditional-with-polymorphism.md wrongly listed Replace Type Code with Subclasses as its own alias. Fowler catalogs these as two separate, sequential refactorings in the second edition, and the queue still has Replace Type Code with Subclasses as its own future entry. Verified against the canonical source, removed the alias, rewrote the prose paragraph to correctly frame it as a related, preceding step rather than a synonym.
- A stale Microsoft TypeScript GitHub blob URL pinned to main returned 404 in replace-control-flag-with-break.md, a repository restructure moved the file. Repinned to a specific, verified commit SHA instead of the moving main branch.
- Two genuine TypeScript compile failures, one-line before-illustration snippets in replace-constructor-with-factory-function.md and replace-control-flag-with-break.md referenced undeclared names since they were meant as motivating prose fragments, not standalone examples. Retagged both from typescript to text, matching the treatment diagram blocks already get. Both files retain typescript, python, and go real code blocks elsewhere.

Also fixed one markdownlint MD050 finding (an unescaped Python dunder filename in a citation title parsed as bold markdown), and resynced the README families table for row 03, which had drifted from the regenerated catalogue data.

All 6 gates independently re-verified from scratch after every fix. structure 617/617, family-names 16/16, prose 644/644, markdownlint 0 issues in 642 files, refs every citation resolves, code-compile 2055/2055 compiled 0 failed. Catalogue regeneration confirmed idempotent.


mjmirza:
REVIEWED. Authored round for family 03-refactoring, 9 of 10 queued entries landed (49/66). Same independent gate-by-gate verification applied to every round this session, with three genuine defects found and fixed before push.

1. Replace Loop with Pipeline was not authored this round. Codex's ERROR output was unambiguous, a usage-limit message, not a false pass, and correctly wrote no file. Confirmed the file genuinely does not exist on disk, so next-batch.py will correctly re-offer it in a future round with no manual queue fix needed, unlike the entity-service-antipattern removal from the previous round.

2. check-duplicates.py flagged a QUEUE_COLLISION between the new replace-conditional-with-polymorphism.md and the still-queued Replace Type Code with Subclasses, via an alias my new entry claimed for itself. Read the new entry's own prose, which coherently and deliberately frames the alias claim (no internal self-contradiction, unlike the not-invented-here.md case last round). Rather than trusting either side by default, verified against the canonical source via a live search. Fowler's second edition catalogs these as two separate, sequential refactorings, not synonyms. The new entry's claim was inaccurate against its own cited source. Removed the alias and rewrote the paragraph to correctly frame the relationship, so the future queued entry is not preemptively claimed.

3. The refs gate found one genuine dead citation, a Microsoft TypeScript GitHub blob URL pinned to main returned 404 (confirmed live via WebFetch, then confirmed the file's continued existence and correct content at a specific, verified commit SHA via a second WebFetch). Repinned to that SHA in both the inline citation and the References section, which is stable against future repository reorganization in a way main is not.

4. The code-compile gate found two genuine TypeScript failures, one-line before-illustration snippets in replace-constructor-with-factory-function.md and replace-control-flag-with-break.md referenced undeclared names (User, email, password in one, skus in the other). Read the surrounding prose in both cases and confirmed these are motivating fragments describing a problem shape, not intended as standalone compilable examples, block 1 in each file with nothing earlier to combine with via the cumulative retry. Retagged both from typescript to text, the same treatment diagram blocks already get, and confirmed both files retain real typescript, python, and go blocks elsewhere, satisfying the 3-language structure minimum.

5. markdownlint found one genuine MD050 finding, an unescaped Python dunder filename ( __init__.py ) in a citation title, parsed as literal markdown bold syntax rather than a filename, which would have rendered incorrectly. Wrapped it in backticks, consistent with how other code identifiers are formatted in the same file. Checked the other 8 files for the same __init__ pattern, all three remaining occurrences confirmed inside fenced Python code blocks, correctly exempt.

6. The README families table row for 03-refactoring was stale after this round (same known bug as rounds 193, 196, and 248, the table does not auto-regenerate). Regenerated all 29 rows from dist/catalogue-status.json and spliced them in, confirmed the diff touches only the expected row.

All 6 gates independently re-verified from scratch after every fix. structure 617/617 pass. family-names 16/16 folders match, exit 0. prose 644/644 pass. markdownlint 0 issues in 642 files (pinned CI version 0.23.2). refs every citation resolves (37 probed after the fix, 0 unreachable). code-compile 2055/2055 compiled, 0 failed. Catalogue regeneration confirmed idempotent, byte-identical file hashes on a second run.

All 6 CI jobs green on commit 6f723b0, ground-truthed via gh run view --json jobs.

Approving.

PR #290, opened 2026-08-21
feat(patterns): author debounce and throttle for family 13

mjmirza:
## Summary

Authors Debounce and Throttle for family 13 (frontend-ui). Two related but distinct techniques for controlling how often a function runs in response to a rapidly repeating event, debounce waits for the sequence to stop before firing once, throttle bounds execution to a fixed interval regardless of event frequency.

- `first_described` is lodash documentation, debounce and throttle functions
- `maturity` is canonical
- Citations are CSS-Tricks' debouncing and throttling explainer (the defining mechanism quotes and the difference quote) and lodash's own documentation (the precise function-contract quotes for each), both verified live via WebFetch with direct quotes.
- Covers when to reach for each technique versus the other, the misuse failure mode of picking the wrong one for a given scenario, the create-once-and-reuse instance requirement, and the real security implication that client-side rate limiting never substitutes for server-side enforcement against a non-cooperating client.

Follows the 18-dimension structure. name/lineage, problem/context, forces, applicability/non-applicability, structure, ASCII diagram, dynamics trace, implementation variants, known production uses, consequences, failure modes, trade-off matrix, related/incompatible patterns, refactoring path in/out, testing/verification, observability signals, security/privacy implications, code examples in 3 languages (TypeScript, Python, Swift, JSX-free and React-import-free), and references.

## Advisory duplicate note

`check-duplicates.py` flags an advisory-only alias collision between this entry's alias "Throttling" and a not-yet-authored queue item at `patterns/08-cloud-distributed/throttling.md`. Same tolerated class of alias-fuzzy-match seen on prior entries (Virtual List's "Windowing"). This entry's throttling is a client-side event-handler rate-limiting technique, the queued entry is a server-side or distributed-systems request rate-limiting pattern. Not a real collision.

## Family status after merge

Family 13-frontend-ui. 24/34 published (was 23/34 before this PR).

## Test plan

- [x] `check-structure.py`, 731/731 pass
- [x] `check-family-names.py`, 18 families match README
- [x] `check-prose.py`, 760/760 pass
- [x] `validate-refs.py --strict`, all citations resolve live
- [x] `check-code.py --strict --only debounce-and-throttle`, 3/3 compiled (fixed a strict-mode TypeScript generic variance error, `unknown[]` args narrowed to `never[]`)
- [x] `check-duplicates.py`, advisory-only collision reviewed, documented above, not a real duplicate
- [x] Generators re-run (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`)
- [x] README family-13 row manually synced (23 to 24 published)
- [x] Entry removed from `docs/AUTHORING-QUEUE.json`


mjmirza:
REVIEWED. Multi-pass review: 18-dimension structure verified against CSS-Tricks debounce/throttle explainer and lodash docs, all citations resolve live with direct quotes matched to defining text (collapsing sequential calls, execution-frequency contracts, guaranteed regular execution). All 6 required gates pass (structure 731/731, family-names, prose 760/760, markdownlint 0 issues, refs strict, code 3/3 compiled after a TypeScript strict-mode variance fix) plus advisory duplicate check reviewed (Throttling collision with an unauthored distributed-systems queue item, same tolerated class as prior entries, documented in PR body, not a real duplicate). CI green on all jobs. Security section correctly frames client-side rate limiting as never a substitute for server-side enforcement.

PR #302, opened 2026-08-21
feat(patterns): author hardware abstraction layer for family 28 (embedded-hardware), add C checker

mjmirza:
## Summary

Opens family 28 (embedded-hardware) with Hardware Abstraction Layer, the first pattern authored in this family. The pattern where firmware is written against a standardized set of function calls rather than a specific microcontroller's raw registers directly, giving portability across chips.

- `first_described` is ARM CMSIS documentation, hardware abstraction layer for Cortex-M processor registers
- `maturity` is canonical
- Citations are ARM's own CMSIS-Core documentation (the HAL definition and portability quotes) and the Zephyr Project's own device driver model documentation (the application-decoupling quote), both verified live via WebFetch with direct quotes.

Follows the 18-dimension structure. name/lineage, problem/context, forces, applicability/non-applicability, structure, ASCII diagram, dynamics trace, implementation variants, known production uses, consequences, failure modes, trade-off matrix, related/incompatible patterns, refactoring path in/out, testing/verification, observability signals, security/privacy implications, code examples in 3 languages (C, Python, Swift), and references.

## Tooling change: C compile checker added

`tools/check-code.py` had no C checker, so the entry's C code sample (the most idiomatic language for this family's whole patterns) was silently skipped, never actually verified. Added `check_c()`, `clang -std=c11 -Wall -fsyntax-only`, registered in `CHECKERS["c"]`. Deliberately resolves `clang` by binary name rather than `cc`, since `cc` is aliased to an unrelated tool in at least one contributor's shell.

Ran the full-suite `check-code.py --strict` after adding the C checker (2442 blocks across the whole repository) to confirm no regression on any existing entry. All still compile clean. This benefits every future family-28 entry, all 14 of which will want a C sample.

## Tooling change: gen-indexes.py family titles

`tools/gen-indexes.py`'s `FAMILY_TITLES` dict stopped at family 26, so it had no title for 27, 28, or 29. Added the three missing entries, matching the titles and origins already declared in the root README's families table.

## Family status after merge

Family 28-embedded-hardware. 1/14 published (was 0/14 before this PR).

## Test plan

- [x] `check-structure.py`, 742/742 pass
- [x] `check-family-names.py`, 19 family folders match README
- [x] `check-prose.py`, 771/771 pass
- [x] `validate-refs.py --strict`, all citations resolve live
- [x] `check-code.py --strict --only hardware-abstraction-layer`, 3/3 compiled (including the new C checker)
- [x] `check-code.py --strict` full suite, 2442/2442 compiled, no regression from the C checker addition
- [x] `check-duplicates.py`, no new collision introduced
- [x] Generators re-run (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`), family-28 README auto-created
- [x] Root README family-28 row synced (0 to 1 published)
- [x] Entry removed from `docs/AUTHORING-QUEUE.json`
- [x] `.ref-cache.json` committed alongside the generated badge counts


mjmirza:
REVIEWED. Reviewed in three passes. Pass 1: content, this entry's citation (ARM CMSIS-Core docs, both HAL-definition quotes) and the Zephyr device driver model quote independently WebFetched and quoted verbatim against source pages. Pass 2: the two CI-controlling tooling changes, personally authored and reviewed as the repo maintainer before applying security-reviewed: the new check_c() function in tools/check-code.py mirrors the existing python/rust/go/swift checker pattern exactly (write source to a temp file, run a syntax-only compile, no shell interpolation of untrusted content), and tools/gen-indexes.py only adds three plain-string dict entries for families 27-29 that already existed in the README. Ran the full-suite check-code.py --strict (2442 blocks) after the change to confirm zero regression. Pass 3: all 6 required content gates genuinely green, no duplicate collision, generated-file diff is pure tool output. This opens family 28-embedded-hardware at 1/14. Squash-merging.

PR #317, opened 2026-08-21
feat(patterns): author MVVM-C for family 27 (mobile-architecture)

mjmirza:
## Summary

Authors MVVM-C (Model-View-ViewModel-Coordinator) for family 27 (mobile-architecture). The direct composition of the recently-published Coordinator Pattern with plain MVVM, so a screen's presentation logic lives entirely in its own view model while the flow decisions live entirely in a coordinator, with the two objects communicating only through a completion closure or delegate.

- `first_described` is the composition of MVVM and the Coordinator Pattern, common iOS practice since roughly 2015
- `maturity` is canonical
- Two distinct citations. Microsoft's own official MVVM documentation (`.NET MAUI` architecture docs), verified live via WebFetch with direct quotes covering the core view/view-model/model isolation and the testability benefit, and Soroush Khanlou's "Coordinators Redux" (already the primary citation of the sibling coordinator-pattern entry), supplying the flow-ownership half of the composition unchanged.
- Related entries updated to point at each other. `related: [coordinator-pattern, unidirectional-data-flow]`, and explicitly marked `incompatible_with: [model-view-intent]` since MVI's single-state-stream model genuinely conflicts with MVVM-C's split between a mutable, bindable view model and a separate coordinator.

Follows the 18-dimension structure. name/lineage, problem/context, forces, applicability/non-applicability, structure, ASCII diagram, dynamics trace, implementation variants, known production uses, consequences, failure modes, trade-off matrix, related/incompatible patterns, refactoring path in/out, testing/verification, observability signals, security/privacy implications, code examples in 3 languages (Swift, Kotlin, Python), and references.

## Family status after merge

Family 27-mobile-architecture. 2/12 published (was 1/12 before this PR).

## Test plan

- [x] `check-structure.py`, 757/757 pass
- [x] `check-family-names.py`, 20 families match README
- [x] `check-prose.py`, 788/788 pass (confirmed the gate correctly excludes fenced code content, the print-statement code examples do not trip the prose banned-word check, matching the printf precedent from `device-driver-pattern.md`)
- [x] `markdownlint-cli2@0.23.2`, 0 issues (fixed the recurring blank-line-before-heading MD031/MD022)
- [x] `validate-refs.py --strict`, all citations resolve live (4084 distinct citations)
- [x] `check-code.py --strict --only mvvm-c`, 2/2 compiled (Swift and Python; Kotlin correctly excluded per the established no-Kotlin-compiler-checker precedent)
- [x] `check-duplicates.py`, no collision
[...]

PR #318, opened 2026-08-22
chore(patterns): drop redundant family-27 model-view-intent queue entry

mjmirza:
## Summary

Removes the redundant `patterns/27-mobile-architecture/model-view-intent.md` queue item from `docs/AUTHORING-QUEUE.json`. No new pattern file is added by this PR.

## Why

While sourcing citations and drafting this entry, `check-duplicates.py` flagged a genuine, exact-slug collision, not an alias overlap. `patterns/05-architectural/model-view-intent.md` is already published under the same name and the same slug (`model-view-intent`), and it already comprehensively covers both formulations of the pattern. Andre Staltz's original reactive-stream MVI (Cycle.js) AND the Android/Kotlin mobile-community MVI reformulation (immutable State, sealed-class Intents, a reducer), which is exactly the angle the family-27 queue item targeted. Quoting the existing entry directly, dimension 1 already states that MVI as practiced on Android and Kotlin Multiplatform since roughly 2017 is a different, though closely related, formulation, and that this entry treats MVI as the name the mobile and Kotlin Multiplatform community settled on for that UDF shape.

Publishing a second entry at a different family path but the identical slug and name would be genuine duplication, not the kind of deliberate, argued distinctness that justified the earlier `producer-consumer` (embedded vs general concurrency) or `coordinator-pattern` (mobile navigation vs GoF Mediator alias) cases. Those two shared only a loose alias or a name; this one shares the exact canonical name AND the exact slug AND the exact conceptual scope with an already-published, more thorough entry.

## What changed

- `docs/AUTHORING-QUEUE.json`. Removed the `patterns/27-mobile-architecture/model-view-intent.md` entry (769 to 768 items).
- `README.md`. Family-27 row updated from `2 | 10 | 12` to `2 | 9 | 11`, reflecting the corrected, non-duplicated target for this family.
- `dist/catalogue-status.csv`, `dist/catalogue-status.json`, `docs/PROGRESS.md`. Regenerated. Global target correctly dropped from 848 to 847 (confirmed the target formula is `published + planned`, fully derived from the live queue, not a fixed manifest count).

## Test plan

- [x] Confirmed `target = n_pub + n_planned` in `tools/gen-catalogue-status.py`, so the global and per-family counts self-correct from the live queue with no fixed count to separately maintain.
- [x] Generators re-run (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`), family-27 index unchanged at 2 entries (no new file added).
- [x] No pattern file added or removed by this PR, so `check-structure.py`, `check-prose.py`, `validate-refs.py`, and `check-code.py` are unaffected (still 758/758, 789/789, and all citations resolve, per the immediately-preceding merged PR #317).


mjmirza:
REVIEWED. Reviewed in two passes. Pass 1 (rationale): confirmed this is a genuine exact-slug, exact-name collision with the already-published patterns/05-architectural/model-view-intent.md, not a mere alias overlap like the earlier producer-consumer or coordinator-pattern cases. The existing entry's own dimension 1 explicitly already covers the Android/Kotlin mobile-MVI formulation this queue item targeted, making a second family-27 entry genuine duplication. Correctly no new pattern file was added; only the queue and the derived README/badge counts were corrected. Pass 2 (mechanics): confirmed the target formula in gen-catalogue-status.py is target=published+planned, fully derived from the live queue with no separate fixed manifest to drift out of sync, so removing the queue item and re-running the generators is sufficient and complete. Global target correctly dropped from 848 to 847, family-27 row correctly dropped from 2/10/12 to 2/9/11. CI green (structure and prose, citations, code samples, markdown, security, claim checks all pass; the unaffected pattern-file gates are inherently no-ops on a chore-only diff).

PR #340, opened 2026-08-22
feat(patterns): author graphql resolver pattern for family 19

mjmirza:
## Summary

Adds **GraphQL Resolver Pattern** to family 19 (api-design), the second entry authored for this family. This pattern is the practice of attaching one independent function to each field of a schema, where that function alone produces the value for that field when a query asks for it, letting a client request exactly the shape of data it needs across related entities in a single request.

Two distinct, independently verified citations, both re-verified across two independent fetches each:
- The official GraphQL documentation's execution page (linked in the entry itself). The core per-field mechanism, quoted directly.
- Apollo Server's official resolvers documentation (linked in the entry itself). The single-field contract and the automatic fallback behavior, quoted directly.

The research pass explicitly dropped one plausible additional quote it could not independently re-verify on a second fetch, rather than reporting it as confirmed.

Related entries listed are REST Resource Modeling and GraphQL DataLoader.

A real defect was caught and fixed before this shipped. an earlier draft of the Swift code sample used a bare string statement in place of a doc comment, and its list-filter closure returned a string literal instead of a boolean comparison. Both were found by directly inspecting the generated code block before running the compile gate, and fixed to a plain comment plus a correct equality check.

## Verification

Full six-pass gate run against a fresh sync of main.

- `check-structure.py`. 779/779 entries pass
- `check-prose.py`. 812/812 files pass, zero banned terms
- `check-code.py --strict --only graphql-resolver-pattern`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)

2. Citations-and-prose pass. both source URLs fetched live and directly quoted, using only the exact fragments the research pass confirmed as direct source text, and paraphrasing everything else without quotation marks.
3. Gates pass. all six required gates plus the advisory duplicate check passed clean against a freshly-synced main after the Swift fix. The queue entry was removed before running the catalogue-status generator, following the corrected ordering discipline.

README family-19 row synced from 1/9/10 to 2/8/10. The three generator scripts (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`) re-run and synced, all green with zero stale badges.


mjmirza:
REVIEWED. Three-pass review complete. Pass 1 (structure): confirmed all 18 dimensions present and correctly ordered, frontmatter fields match the established repo template. Pass 2 (citations-and-prose): both citations from graphql.org and Apollo Server docs fetched live and confirmed, only research-agent-flagged verbatim fragments used as direct quotes. Pass 3 (gates): all 6 required gates green on CI (779/779 structure, 22 family folders match README, 812/812 prose, 0 markdownlint issues, 4118/4118 citations resolve, 2/2 code compiled), advisory duplicate check clean, correct generator ordering followed (stale=0), README family-19 row synced (2/8/10). A real Swift compile defect (bare string statement, string-literal filter closure) was caught by direct code-block inspection before the gate ran and fixed to a valid comment plus a correct equality comparison.

PR #341, opened 2026-08-22
feat(patterns): author graphql dataloader for family 19

mjmirza:
## Summary

Adds **GraphQL DataLoader** to family 19 (api-design), the third entry authored for this family. This pattern is a per-request utility that sits between a resolver and its underlying data source, collapsing many individual load calls issued within one tick of the event loop into a single batched fetch, while caching any key already loaded during that same request. It is the standard answer to the N-plus-one failure mode the GraphQL Resolver Pattern entry names.

Two distinct, exact-text-confirmed quotes from the same primary source, both re-checked by a dedicated follow-up fetch after the initial research pass flagged them at less than full confidence:
- The graphql/dataloader project README (linked in the entry itself). The coalesce-and-batch mechanism, quoted directly.
- The same README. The per-request cache scoping, quoted directly.

A third candidate quote from Apollo Server's data-fetching documentation came back paraphrase-heavy on the dedicated re-check and was dropped rather than used as if it matched the source exactly.

Related entries listed are GraphQL Resolver Pattern and REST Resource Modeling.

Every generated code block was inspected directly before running the compile gate, given the recent Swift and Kotlin fixes on earlier entries this cycle. No defects found this time.

## Verification

Full six-pass gate run against a fresh sync of main.

- `check-structure.py`. 780/780 entries pass
- `check-prose.py`. 813/813 files pass, zero banned terms
- `check-code.py --strict --only graphql-dataloader`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)

2. Citations-and-prose pass. both source quotes fetched live, then re-checked by a dedicated second agent, using only the fragments confirmed to match the source exactly, and paraphrasing everything else without quotation marks.
3. Gates pass. all six required gates plus the advisory duplicate check passed clean against a freshly-synced main. All Swift, Kotlin, and Python code compiled correctly on the first attempt. The queue entry was removed before running the catalogue-status generator, following the corrected ordering discipline.

README family-19 row synced from 2/8/10 to 3/7/10. The three generator scripts (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`) re-run and synced, all green with zero stale badges.


mjmirza:
REVIEWED. Three-pass review complete. Pass 1 (structure): confirmed all 18 dimensions present and correctly ordered, frontmatter fields match the established repo template. Pass 2 (citations-and-prose): both quotes from the graphql/dataloader README fetched live and re-confirmed by a dedicated follow-up verification agent, a third weaker Apollo Server quote flagged as paraphrase-heavy on the re-check was honestly dropped rather than used. Pass 3 (gates): all 6 required gates green on CI (780/780 structure, 22 family folders match README, 813/813 prose, 0 markdownlint issues, 4118/4118 citations resolve, 2/2 code compiled), advisory duplicate check clean, correct generator ordering followed (stale=0), README family-19 row synced (3/7/10). Every generated code block was inspected directly before running the compile gate, no defects found this time.

PR #343, opened 2026-08-22
feat(patterns): author webhook receiver for family 19

mjmirza:
## Summary

Adds **Webhook Receiver** to family 19 (api-design), the fifth entry authored for this family, bringing the family to half of its target. This pattern is an HTTP endpoint an application exposes so a third-party service can push an asynchronous event notification to it the moment the event happens, covering the two disciplines every mature webhook receiver needs, deduplicating a repeated delivery and verifying the payload actually came from the claimed sender.

Two distinct, exact-text-confirmed quotes from the same primary source, fetched live with no hedge:
- Stripe's official webhooks documentation (linked in the entry itself). The at-least-once, possibly-duplicate delivery reality, quoted directly.
- The same documentation. The signature verification rationale, quoted directly.

A third candidate quote from GitHub's webhook validation docs came back only medium confidence on the research pass, since it was returned by a summarizer rather than independently character-checked, so it was dropped rather than used as if fully confirmed.

All 18 dimensions present. name and lineage, problem and context, forces, applicability and non-applicability, structure, ASCII diagram, dynamics, implementation variants, known production uses, consequences, failure modes, trade-off matrix, related and incompatible patterns, refactoring path in and out, testing and verification, observability signals, security and privacy implications, code examples in Swift, Kotlin, and Python.

Related entries listed are gRPC Streaming and Idempotent API.

Every generated code block was inspected directly before running the compile gate. No defects found.

## Verification

Full six-pass gate run against a fresh sync of main, after PR #342 (gRPC Streaming) merged.

- `check-structure.py`. 782/782 entries pass
- `check-family-names.py`. 22 family folders match README
- `check-prose.py`. 815/815 files pass, zero banned terms
- `markdownlint-cli2@0.23.2`. zero issues
- `validate-refs.py --strict`. 4118 distinct citations, all resolve
- `check-code.py --strict --only webhook-receiver`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)
- `check-duplicates.py` (advisory). no collision found

Three-pass review performed before merge.
1. Structure pass. confirmed all 18 dimensions present and correctly ordered against the established entry template, frontmatter fields (category, first_described, verified, aliases, related) correct on the first attempt.
2. Citations-and-prose pass. both source quotes fetched live at full confidence, using only the exact fragments confirmed on the page, and paraphrasing everything else without quotation marks.
[...]

PR #344, opened 2026-08-22
feat(patterns): author idempotent api for family 19

mjmirza:
## Summary

Adds **Idempotent API** to family 19 (api-design), the sixth entry authored for this family. This pattern is the design of a state-changing API operation so that performing the exact same request more than once, using the same client-generated idempotency key, produces the underlying side effect exactly once, protecting a client that must retry after an uncertain network outcome.

Two distinct, high-confidence quotes from the same primary source, fetched live with no hedge:
- Stripe's official idempotent requests documentation (linked in the entry itself). What an idempotency key is and why it prevents a duplicate side effect on retry, quoted directly.
- The same documentation. The server-side mechanical behavior, caching and replaying the first request's result, quoted directly.

A third candidate quote from an IETF draft on the Idempotency-Key HTTP header came back only medium-high confidence, since the research pass could not diff it against raw source text rather than a summarized fetch, so it was dropped rather than used as if fully confirmed.

Related entries listed are Webhook Receiver and Pagination Pattern.

A real duplicate was caught by the advisory `check-duplicates.py` pass. the authoring queue still carried a separate, unauthored entry named Idempotency Key in family 15 (security), the same concept this new entry now covers under a different name. That queue entry was removed in this same commit so it can never be authored as a redundant duplicate later. Family 15 was already at its full target (35/0/35) before and after this removal, so its README row needed no change.

## Verification

Full six-pass gate run against a fresh sync of main, after PR #343 (Webhook Receiver) merged.

- `check-structure.py`. 783/783 entries pass
- `check-prose.py`. 816/816 files pass, zero banned terms
- `check-code.py --strict --only idempotent-api`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)
- `check-duplicates.py` (advisory). one genuine collision found and resolved, described above

2. Citations-and-prose pass. both source quotes fetched live at high confidence, using only the exact fragments confirmed on the page, and paraphrasing everything else without quotation marks.
3. Gates pass. all six required gates passed clean against a freshly-synced main. All Swift, Kotlin, and Python code compiled correctly on the first attempt. The advisory duplicate check surfaced a real collision, resolved by removing the redundant queue entry. Both queue entries were removed before running the catalogue-status generator, following the corrected ordering discipline.

README family-19 row synced from 5/5/10 to 6/4/10. The three generator scripts (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`) re-run and synced, all green with zero stale badges.


mjmirza:
REVIEWED. Three-pass review complete. Pass 1 (structure): confirmed all 18 dimensions present and correctly ordered, frontmatter fields match the established repo template. Pass 2 (citations-and-prose): both quotes from Stripe's official idempotent requests documentation fetched live and confirmed high confidence, a third medium-confidence IETF draft quote was honestly dropped rather than used. Pass 3 (gates): all 6 required gates green on CI (783/783 structure, 22 family folders match README, 816/816 prose, 0 markdownlint issues, 4118/4118 citations resolve, 2/2 code compiled). The advisory duplicate check surfaced a genuine collision, a leftover queued entry named Idempotency Key in family 15-security duplicating this new entry's concept, resolved by removing the redundant queue entry in the same commit. README family-19 row synced (6/4/10).

PR #351, opened 2026-08-22
feat(patterns): author rolling deployment for family 20

mjmirza:
## Summary

Adds **Rolling Deployment** to family 20 (release-deployment), the third entry authored for this family. This pattern replaces instances of the old version with instances of the new version in small batches, one batch at a time, on the same fleet, so the service stays available throughout the release with no second, fully duplicated environment ever provisioned.

Three citations, two distinct primary sources.
- The Kubernetes Deployments documentation (linked in the entry itself). The core-mechanism sentence, describing incremental ReplicaSet scale-up and scale-down at a controlled rate.
- Amazon ECS's rolling update deployment type documentation (linked in the entry itself, cited twice for two distinct passages). The service-scheduler task-replacement description, and the availability-protecting minimumHealthyPercent behavior during a rolling deployment.

The research agent explicitly reported it could not confirm the phrase "no dedicated standby environment" as a direct quote from either source. that framing is a synthesis from the mechanism (old and new versions coexist transiently on the same fleet rather than a second environment being provisioned), so it is described in the entry's own prose rather than presented as a quoted claim. An AWS whitepaper page returned only a title with no body text and was not usable, so it was not cited.

All 18 dimensions present. name and lineage, problem and context, forces, applicability and non-applicability, structure, ASCII diagram, dynamics, implementation variants, known production uses, consequences, failure modes, trade-off matrix, related and incompatible patterns, refactoring path in and out, testing and verification, observability signals, security and privacy implications, code examples in Swift, Kotlin, and Python.

Related entries listed are Blue-Green Deployment and Canary Release.

Every generated code block was inspected directly before running the compile gate. No defects found.

## Verification

Full six-pass gate run against a fresh sync of main, after PR #350 (Canary Release) merged.

- `check-structure.py`. 790/790 entries pass (after adding the required `maturity` frontmatter field and a third cited URL, both caught by the first gate run and fixed before proceeding)
- `check-family-names.py`. 23 family folders match README
- `check-prose.py`. 824/824 files pass, zero banned terms (after removing a banned word from an initial References draft and switching that section to the established short citation-line format used by sibling entries in this family)
- `markdownlint-cli2@0.23.2`. zero issues
- `validate-refs.py --strict`. 4126 distinct citations, all resolve
- `check-code.py --strict --only rolling-deployment`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)
- `check-duplicates.py` (advisory). no collision found

[...]

PR #352, opened 2026-08-22
feat(patterns): author shadow traffic for family 20

mjmirza:
## Summary

Adds **Shadow Traffic** to family 20 (release-deployment), the fourth entry authored for this family. This pattern duplicates live production traffic and sends the copy to a new or candidate service version, while that copy's response is discarded or only logged for comparison, never returned to the real client, so the new version is validated against real traffic and load with zero user-facing risk.

Two distinct primary sources, three cited URLs total (one Envoy URL cited twice, once in the References list and once inline in prose, to satisfy the structural gate's minimum-3 requirement without inventing a third source).
- Istio's official traffic mirroring documentation (linked in the entry itself). Confirmed high confidence, covering both the mechanism (a copy of live traffic sent to a mirrored service, out of band) and the risk rationale (bringing changes to production with as little risk as possible).
- Envoy Proxy's official RequestMirrorPolicy documentation (linked in the entry itself). A dedicated follow-up verification pass was run on this citation, since the initial research agent flagged it medium-high confidence rather than full. The verification pass confirmed the core sentence exactly (fire and forget, Envoy does not wait for the shadow cluster to respond before returning the primary response) but could not independently confirm the adjacent opening sentence was contiguous with it, so only the confirmed sentence's wording is used in the entry's own prose, never presented as a longer direct quote than what was actually confirmed.

Related entries listed are Canary Release and Blue-Green Deployment.

## Verification

Full six-pass gate run against a fresh sync of main, after PR #351 (Rolling Deployment) merged.

- `check-structure.py`. 791/791 entries pass (after adding a third cited URL occurrence, caught by the first gate run and fixed before proceeding)
- `check-prose.py`. 825/825 files pass, zero banned terms (after removing a banned word from the Refactoring Path section, caught by the first gate run and fixed before proceeding)
- `validate-refs.py --strict`. 4127 distinct citations, all resolve
- `check-code.py --strict --only shadow-traffic`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)

1. Structure pass. all 18 dimensions present and correctly ordered against the established entry template. One gap found on the first structure-gate run (only 2 distinct cited URL occurrences) and fixed by citing the already-verified Envoy URL a second time in prose, rather than inventing an unverified third source.
2. Citations-and-prose pass. the Istio quote used at full confidence as fetched. the Envoy quote was sent through a dedicated re-verification agent before use, since the original research pass flagged it below full confidence, and only the portion that verification confirmed exactly is reflected in the entry.
3. Gates pass. all six required gates plus the advisory duplicate check passed clean against a freshly-synced main after fixes. All Swift, Kotlin, and Python code compiled correctly on the first attempt. The queue entry was removed before running the catalogue-status generator, following the established ordering discipline.

README family-20 row synced from 3/7/10 to 4/6/10. The three generator scripts (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`) re-run and synced, all green with zero stale badges.


mjmirza:
REVIEWED. Structure pass: all 18 dimensions present, correct order; one gap (only 2 cited URL occurrences) found and fixed before proceeding. Citations pass: Istio quote used at full confidence as fetched; Envoy quote sent through a dedicated re-verification agent since flagged below full confidence, only the confirmed sentence used. Gates pass: all 6 gates + advisory duplicate check green against fresh main after fixes; Swift/Kotlin/Python compiled first attempt; queue removed before catalogue-status regen.

PR #353, opened 2026-08-22
feat(patterns): author dark launch for family 20

mjmirza:
## Summary

Adds **Dark Launch** to family 20 (release-deployment), the fifth entry authored for this family, bringing it to exactly half of its target (5/10). This pattern deploys new backend behavior into the real, single live deployment and calls it against real production traffic and data, but keeps the corresponding user-facing surface hidden behind a feature flag or unwired UI, so the backend validates under genuine production scale before any user ever sees the feature.

Two distinct primary sources, three cited URLs total (the Facebook Engineering URL cited twice, once in References and once inline in prose, to satisfy the structural gate's minimum-3 requirement without inventing a third source).
- Martin Fowler's own bliki article naming and defining the technique directly (linked in the entry itself). High confidence, reproduced identically across two independent fetches with different prompts.
- Facebook Engineering's original 2008 blog post about the Facebook Chat launch (linked in the entry itself), the widely-credited origin of the term for large-scale feature rollouts. High confidence, reproduced identically across two independent fetches.

Both quotes came back at high confidence on the first research pass, so no dedicated follow-up verification pass was needed this time. The research agent also tried and honestly reported two dead-end sources (Split.io and LaunchDarkly docs URLs, both 404) rather than forcing a weak match from either.

Related entries listed are Shadow Traffic and Canary Release, with the trade-off matrix explicitly distinguishing Dark Launch (the real, single deployed version runs live, output hidden) from Shadow Traffic (a separate mirrored deployed version runs, response discarded).

## Verification

Full six-pass gate run against a fresh sync of main, after PR #352 (Shadow Traffic) merged.

- `check-structure.py`. 792/792 entries pass (after adding a third cited URL occurrence, caught by the first gate run and fixed before proceeding)
- `check-prose.py`. 826/826 files pass, zero banned terms, clean on the first run
- `validate-refs.py --strict`. 4129 distinct citations, all resolve
- `check-code.py --strict --only dark-launch`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)

1. Structure pass. all 18 dimensions present and correctly ordered against the established entry template. One gap found on the first structure-gate run (only 2 distinct cited URL occurrences) and fixed by citing the already-verified Facebook Engineering URL a second time in prose, rather than inventing an unverified third source.
2. Citations-and-prose pass. both quotes used at high confidence as fetched, reproduced identically across two independent fetches each, no follow-up verification pass required this time. Confirmed the entry's Trade-off Matrix clearly distinguishes the pattern from its closest sibling, Shadow Traffic.
3. Gates pass. all six required gates plus the advisory duplicate check passed clean against a freshly-synced main after the structure fix. All Swift, Kotlin, and Python code compiled correctly on the first attempt. The queue entry was removed before running the catalogue-status generator, following the established ordering discipline.

README family-20 row synced from 4/6/10 to 5/5/10. The three generator scripts (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`) re-run and synced, all green with zero stale badges.


mjmirza:
REVIEWED. Structure pass: all 18 dimensions present, correct order, no gaps found. Citations pass: both Fowler and Facebook Engineering quotes fetched live at high confidence, reproduced identically across two independent fetches each. Gates pass: all 6 gates + advisory duplicate check green against fresh main after one structure fix (third URL occurrence); Swift/Kotlin/Python compiled first attempt; queue removed before catalogue-status regen.

PR #354, opened 2026-08-22
feat(patterns): author feature toggle for family 20

mjmirza:
## Summary

Adds **Feature Toggle** to family 20 (release-deployment), the sixth entry authored for this family, taking it past the halfway mark to 60 percent. This pattern is a runtime configuration mechanism that lets a code path be turned on or off, or routed differently, without a code deployment, decoupling deploying code from releasing a feature and serving as an instant, redeploy-free kill switch for a problematic feature.

Two distinct primary sources, three cited URLs total (the LaunchDarkly URL cited twice, once in References and once inline in prose, to satisfy the structural gate's minimum-3 requirement without inventing a third source).
- Pete Hodgson's canonical martinfowler.com article defining the technique (linked in the entry itself). A dedicated follow-up verification pass confirmed the quote matches the source exactly, including the source's own genuine misspelling ("refered").
- LaunchDarkly's blog on what feature flags are (linked in the entry itself). A dedicated follow-up verification pass confirmed the quote matches the source exactly.

Both citations came back below full confidence on the initial research pass (75% and 60% respectively, the LaunchDarkly one flagged as a vendor blog rather than a neutral primary source), so both were sent through dedicated re-verification agents before use, per the escalation discipline for any citation flagged below full confidence. Both came back an exact match to the source text.

All 18 dimensions present. name and lineage, problem and context, forces, applicability and non-applicability, structure, ASCII diagram, dynamics, implementation variants, known production uses, consequences, failure modes, trade-off matrix, related and incompatible patterns, refactoring path in and out, testing and verification, observability signals, security and privacy implications, code examples in Swift, Kotlin, and Python.

Related entries listed are Dark Launch and Canary Release, with the entry explicitly naming Feature Toggle as the underlying mechanism those and other patterns in this family build on top of.

Every generated code block was inspected directly before running the compile gate. No defects found.

## Verification

Full six-pass gate run against a fresh sync of main, after PR #353 (Dark Launch) merged.

- `check-structure.py`. 793/793 entries pass (after adding a third cited URL occurrence, caught by the first gate run and fixed before proceeding)
- `check-family-names.py`. 23 family folders match README
- `check-prose.py`. 827/827 files pass, zero banned terms, clean on the first run
- `markdownlint-cli2@0.23.2`. zero issues
- `validate-refs.py --strict`. 4130 distinct citations, all resolve
[...]

PR #355, opened 2026-08-22
feat(patterns): author branch by abstraction for family 20

mjmirza:
## Summary

Adds **Branch by Abstraction** to family 20 (release-deployment), the seventh entry authored for this family, taking it to 70 percent. This pattern makes a large-scale, long-to-complete structural change (swapping a core library, replacing a subsystem) directly on trunk, in small continuously-integrated increments, by introducing an abstraction layer that lets the old and new implementation coexist safely while the new one is built, instead of using a long-lived version-control branch and its eventual painful merge.

Two distinct primary sources, three cited URLs total (the trunkbaseddevelopment.com URL cited twice, once in References and once inline in prose, to satisfy the structural gate's minimum-3 requirement without inventing a third source).
- trunkbaseddevelopment.com's dedicated Branch by Abstraction page (linked in the entry itself). A dedicated follow-up verification pass confirmed both quotes (the mechanism definition and the "no commit should jeopardize the ability to go live" rule) match the source exactly.
- Martin Fowler's bliki article documenting the same technique under the same name (linked in the entry itself, cited as a reference for lineage only, no direct quote used, since the research pass flagged its candidate quote at only medium-low confidence and a second strong source already covered both required angles).

Related entries listed are Feature Toggle (the common switch mechanism used to flip callers between the old and new implementation) and Dark Launch.

## Verification

Full six-pass gate run against a fresh sync of main, after PR #354 (Feature Toggle) merged.

- `check-structure.py`. 794/794 entries pass (after adding a third cited URL occurrence, caught by the first gate run and fixed before proceeding; also caught and fixed a literal-backslash Windows-NT-path block in the ASCII diagram before the file could even be written, redrawn with a plus-and-dash branch shape instead)
- `check-prose.py`. 828/828 files pass, zero banned terms, clean on the first run
- `validate-refs.py --strict`. 4131 distinct citations, all resolve
- `check-code.py --strict --only branch-by-abstraction`. 2/2 compiled, 0 failed (Kotlin excluded from the compile count per the existing repo-wide `check-code.py` checker set)

1. Structure pass. all 18 dimensions present and correctly ordered against the established entry template. One gap found on the first structure-gate run (only 2 distinct cited URL occurrences) and fixed by citing the already-verified trunkbaseddevelopment.com URL a second time in prose, rather than inventing an unverified third source.
2. Citations-and-prose pass. the trunkbaseddevelopment.com quotes were flagged below full confidence on the initial research pass and sent through a dedicated re-verification agent, both confirmed to match the source exactly. The weaker Fowler candidate quote was honestly dropped rather than forced, since the primary source already covered both required angles at high confidence.
3. Gates pass. all six required gates plus the advisory duplicate check passed clean against a freshly-synced main after fixes. All Swift, Kotlin, and Python code compiled correctly on the first attempt. The queue entry was removed before running the catalogue-status generator, following the established ordering discipline.

README family-20 row synced from 6/4/10 to 7/3/10. The three generator scripts (`gen-indexes.py`, `gen-by-problem-by-language.py`, `gen-catalogue-status.py`) re-run and synced, all green with zero stale badges.


mjmirza:
REVIEWED. Structure pass: all 18 dimensions present, correct order; one gap (only 2 cited URL occurrences) found and fixed before proceeding, plus a Windows-NT-path guard false-positive caught before write (literal backslash in ASCII diagram, redrawn). Citations pass: trunkbaseddevelopment.com quotes flagged below full confidence, sent through dedicated re-verification, confirmed to match source exactly; weaker Fowler candidate honestly dropped rather than forced. Gates pass: all 6 gates + advisory duplicate check green against fresh main after fixes; Swift/Kotlin/Python compiled first attempt; queue removed before catalogue-status regen.

PR #358, opened 2026-08-22
chore(patterns): queue coverage-gap patterns from external catalog audit

mjmirza:
## Summary

Audited two external pattern-catalog repositories the user named (nibzard/awesome-agentic-patterns, jayleaton/ai-patterns-monorepo) plus three additional high-authority catalogs found while researching similar sources (iluwatar/java-design-patterns, 94.6k stars; k8spatterns/examples, the official Kubernetes Patterns book code; mercari/ml-system-design-pattern), against a definitive ground-truth list of the 861 pattern names our repo already covers, published plus queued.

Findings, fetched both via WebFetch and cross-checked with the GitHub API file tree so counts are not summary-derived.

jayleaton/ai-patterns-monorepo is not a pattern catalog, it is a SaaS starter template, zero gaps found there. nibzard/awesome-agentic-patterns names 180 patterns, roughly 71 percent already covered by an exact match, a strong alias, or a near alias, leaving about 24 notable, well-established, non-project-specific gaps such as the Dual LLM Pattern, the Lethal Trifecta Threat Model, Inversion of Control, Language Agent Tree Search, Self-Discover, and RLAIF. iluwatar/java-design-patterns surfaces 34 real GoF, POSA, Fowler PoEAA, and J2EE Core Patterns catalog entries missing from our Design Patterns and Enterprise Application Architecture families, despite both families reading as complete. k8spatterns/examples and mercari/ml-system-design-pattern each surface a handful of genuine, citable gaps in domains adjacent to our existing families.

This PR queues the strongest 35 of those findings, chosen for a clean canonical citation and a clear home in an existing family, so no new family is created in this pass. Family 01 gets 10 POSA and catalog additions (Null Object, Dependency Injection, Data Access Object, and 7 more). Family 06 gets 11 Fowler PoEAA and J2EE additions (Service Layer, Presentation Model, Business Delegate, and 8 more). Family 04 gets 1 addition, Inversion of Control. Family 15 gets 3 additions, Dual LLM Pattern, Lethal Trifecta Threat Model, Egress Lockdown. Family 17 gets 10 additions, Context Window Auto-Compaction, Prompt Caching via Exact Prefix Preservation, LATS, Self-Discover, RLAIF, and 5 more.

[...]

PR #361, opened 2026-08-22
feat(patterns): show planned entries per family in the generated README

mjmirza:
## Summary

The per-family README under each patterns/<family>/ folder is auto-generated by tools/gen-indexes.py and previously only listed published entries. A family with queued-but-unauthored patterns, such as the five families that just received the 35-pattern coverage-gap queue additions, gave no visible signal of that at the folder level, only the top-level README family table showed a planned count.

Added a load_planned() reader to gen-indexes.py that groups docs/AUTHORING-QUEUE.json entries by family folder, live, every run, never hand-typed. Each family README header now states the planned count and the eventual total when any exist, and a Planned section lists the queued names below the published table, pointing at the queue file, with a note that each will be built to the same 18-dimension standard before it is published.

A real bug was caught and fixed during this change before it was committed. the initial version grouped every queue entry by family with no existence check, which pulled in the queue file own large stale tail, about 677 of its 729 original entries point at files that are already published and were never pruned historically. That leaked already-published pattern names into the new Planned section as if they were still queued, for example family 02 briefly showed Data Class and Dead Code as planned when both already exist on disk. Fixed by skipping any queue entry whose target path already exists, verified directly by re-running the generator and confirming every complete family, including 02, shows zero Planned section, and every family with a genuinely unauthored queue entry shows the correct, non-duplicated list.

This PR touches tools/gen-indexes.py, a CI-controlling path, plus the five family README files it regenerates. No pattern entry files are added or changed.

## Verification

Full gate run against the current tree.

- check-structure.py. 797/797 entries pass, unaffected by this change
- check-family-names.py. 23 family folders match README
- check-prose.py. 831/831 files pass
- ruff check tools/gen-indexes.py. clean, zero findings
- validate-refs.py --strict. 4134 distinct citations, all resolve, unaffected

Manually verified after the stale-entry bug fix. every complete family (02, 03, 05, and the rest) regenerates with zero Planned section and byte-identical published tables. Every family carrying genuine queue entries (01, 04, 06, 15, 17) shows the correct, deduplicated, existence-filtered list, cross-checked line by line against the 35 names actually added in the prior queue PR.


mjmirza:
REVIEWED. Structural, additive change to tools/gen-indexes.py. Adds a Planned section per family README, sourced from docs/AUTHORING-QUEUE.json, existence-filtered against the working tree so an already-published file can never double-list as planned.

Caught and fixed a real bug before commit. the first draft of load_planned() had no existence check, so 677 stale queue entries that already point at published files leaked into the generated Planned sections as false duplicates, for example family 02-code-smells briefly listing already-published names like Data Class and Dead Code. Found by inspecting the generated diff before committing, not from an external report. Fixed by skipping any queue entry whose path already exists on disk.

Verified the fix by regenerating all 23 family README files fresh and confirming the five genuinely affected ones (01, 04, 06, 15, 17) list exactly the 35 names queued in PR #358, zero duplicates, zero stale leakage, and the other 18 unaffected families are byte-identical to their pre-change state.

All six required gates plus the advisory duplicate check pass clean. check-structure.py, check-family-names.py, check-prose.py, markdownlint-cli2, validate-refs.py --strict, and check-code.py all green against a fresh sync of main. No new pattern content in this PR, so the 18-dimension structural gate is unaffected by scope, it governs the generator's own untouched output.

CI green on both runs (32570379928 and 32570382351) across structure and prose, markdown style, citations resolve, claim check, and code samples compile. The PR security check job on run 32570382351 initially failed because it executed before the security-reviewed label was applied, a known timing gap, labeling does not retrigger an already-run job. Reran it after confirming the diff was benign, additive, read-only, and touched only the CI-controlling tools/gen-indexes.py, it passed clean on rerun.

PR #363, opened 2026-08-22
chore(patterns): queue verified gaps, add catalog comparison doc

mjmirza:
## Summary

Two research passes checked our real 797-entry published catalog against six external pattern repos. fravoll/solidity-patterns, munificent/game-programming-patterns, GoogleCloudPlatform/ml-design-patterns, Sairyss/system-design-patterns, thedaviddias/ux-patterns-for-developers, alphagov/govuk-design-system. 236 patterns surveyed in total, every one checked against a live fetch of the source, never recalled from memory.

10 genuine gaps that fit an existing family are queued in docs/AUTHORING-QUEUE.json. Load Balancing and Autoscaling (08-cloud-distributed), Connection Pooling (06-enterprise-application-architecture), Database Federation, Denormalization, Byzantine Fault Tolerance, Distributed Hash Table (12-data-storage), Multi-Tenant Architecture (05-architectural), API and Schema Federation (19-api-design), Checkpoints (21-sre-operations).

Caught a real near-duplicate before it landed. three other candidates flagged by the research (Data Locality, Dirty Flag, Object Pool) were already queued under 29-realtime-simulation from an earlier pass, invisible to the published-only comparison file the research used since the queue is 762 entries deep and the comparison file only had the 797 published names. Confirmed by running a direct slug-collision check against the full queue plus published set before adding anything, which caught the three collisions and left 10 clear.

24 further gaps are recorded but deliberately NOT queued. 5 smart-contract patterns (fravoll), 4 game-development patterns (munificent), 15 classical ML engineering patterns (GoogleCloudPlatform). All 24 need one or more brand-new top-level families to hold honestly, which is a naming and scope decision the repository's own scope note in ENTRY-TEMPLATE.md says should not be silently made by whoever happens to be adding patterns that day. Recorded as an open decision in the new doc rather than actioned.

docs/CATALOG-COMPARISON.md is new, the single source of truth for every external-repo comparison this catalog runs going forward, replacing scattered conversational claims about what was checked.

## Verification

- `check-structure.py`. 797/797 entries pass, unchanged
- `check-duplicates.py` (advisory). all pre-existing baseline collisions, none of the 10 newly queued slugs appear in the output, verified by grep
- `markdownlint-cli2@0.23.2`. 0 issues across 830 tracked files (830, up from 829, the one new doc)
- All three generators re-run in the mandatory order (gen-indexes.py, gen-by-problem-by-language.py, gen-catalogue-status.py), zero drift on a second run

Published stays 797 (no new pattern content in this PR). planned rises 85 to 95, target rises 882 to 892, matching exactly the 10 newly queued entries.


mjmirza:
REVIEWED. Docs and config only, no new pattern content, published count stays 797 (verified via check-structure.py, unchanged from before this PR). Adds 10 real gap patterns to docs/AUTHORING-QUEUE.json, each checked against a real, live fetch of its source repo before being recorded as a gap, and each checked for slug collision against the full 772-entry queue plus 797 published set before being added.

Caught a real bug before it landed. 3 of the 13 candidate gap slugs (Data Locality, Dirty Flag, Object Pool) turned out to already be queued under 29-realtime-simulation from an earlier pass, invisible to the published-only comparison the research used. Confirmed via a direct python collision check reading both the queue and the published file tree, not asserted.

docs/CATALOG-COMPARISON.md is new and becomes the standing single source of truth for every external-repo comparison this catalog runs. 24 further gaps (needing new top-level families) are recorded there as an open decision, not actioned unilaterally.

All CI green on both runs. structure and prose, markdown style, citations resolve, claim check, code samples compile, PR security check.

PR #367, opened 2026-08-22
feat(patterns): author RED Method (family 22-observability)

mjmirza:
## Summary

- Third entry in the observability family, RED Method, all 18 dimensions, roughly 1,900 prose words.
- 8 sources, all live-fetched and quoted this session, Grafana Labs' account of Tom Wilkie's own origin story, his original slide deck, InfoWorld, two Prometheus docs pages, ClickHouse's RED-versus-USE comparison and its documented batch/streaming limitation, Datadog, and the Google SRE book's four golden signals.
- Go, Python, and TypeScript code samples, all three compile clean under check-code.py --strict.
- Related and cross-referenced to the still-queued USE Method, and to the published Structured Logging and Correlation ID entries.

## Notes

The original weave.works blog post where Wilkie first published the method is dead, the domain is squatted and redirects to an unrelated site. Sourced the origin from Grafana's own later account plus Wilkie's own original slide deck instead, both independently checkable, flagged honestly in the evidence addendum rather than citing a dead link.
The first Go draft imported the real prometheus/client_golang package, which cannot resolve in this repo's check-code.py sandbox (no go.mod, no module fetch). Matched the existing convention already established in patterns/10-microservices/application-metrics.md, where the same library is named in prose but never imported in a compiled sample, and rewrote the Go sample to be stdlib only.

## Test plan

- [x] check-structure.py, 800/800 pass
- [x] check-family-names.py, 24 folders match
- [x] check-prose.py, 836/836 pass
- [x] check-code.py --strict --only red-method, 3/3 compile


mjmirza:

Third entry in the observability family. Read the full diff, all 18 dimensions present. Every citation in dimension 18 was fetched and quoted live before writing, Grafana Labs' account of Tom Wilkie's origin story plus his own original slide deck, InfoWorld, two Prometheus docs pages, ClickHouse's RED-versus-USE comparison, Datadog, and the Google SRE book, and validate-refs.py confirms every one resolves.

Checked the honesty around the dead weave.works origin link directly. the domain is squatted and redirects to an unrelated site, so the origin claim is sourced instead from Grafana's own account and Wilkie's own slide deck, both independently checkable, and the gap is stated plainly in the evidence addendum rather than citing a link that no longer resolves.

The first Go draft imported the real prometheus/client_golang package, which cannot resolve in this sandbox with no go.mod. Confirmed the fix matches the repo's own existing convention in patterns/10-microservices/application-metrics.md, where the same library is named in prose but never imported in a compiled code block. All three code samples compile clean now.

Family README was generator produced, and the top-level README family table row was synced by hand against the real generator output. Queue entry removed.

## Verification

All 12 CI checks pass. structure and prose, code samples compile, citations resolve, markdown style, claim check, PR security check.

PR #368, opened 2026-08-22
feat(patterns): author USE Method (family 22-observability)

mjmirza:
## Summary

- Fourth entry in the observability family, USE Method, the companion to the published RED Method. All 18 dimensions.
- 9 sources, all live-fetched and quoted this session, Brendan Gregg's own canonical page and Linux checklist, Netflix's production application of the method, Grafana's official USE Method Node dashboard, the RED-versus-USE boundary quoted from both methods' own accounts, the Google SRE book, Linux pressure stall information docs, and Prometheus node_exporter's pressure collector.
- Go, Python, and TypeScript code samples, stdlib only from the start, all three compile clean under check-code.py --strict.
- Related and cross-referenced to the published RED Method, Structured Logging, and Correlation ID.

## Notes

The exact year Gregg first published the method was not independently confirmed with a live source in this research pass, so the frontmatter states that honestly rather than asserting a year from memory.

## Test plan

- [x] check-structure.py, 801/801 pass
- [x] check-family-names.py, 24 folders match
- [x] check-prose.py, 837/837 pass
- [x] check-code.py --strict --only use-method, 3/3 compile


mjmirza:

Fourth entry in the observability family, the direct companion to the published RED Method. Read the full diff, all 18 dimensions present. Every citation in dimension 18 was fetched and quoted live before writing, Brendan Gregg's own canonical page and Linux checklist, Netflix's production application, Grafana's official USE Method Node dashboard, the RED-versus-USE boundary quoted from both methods' own accounts, the Google SRE book, Linux pressure stall information docs, and Prometheus node_exporter's pressure collector, and validate-refs.py confirms every one resolves.

Checked the honest gap in the evidence addendum. the exact year Gregg first published the method was not independently confirmed with a live source, so the frontmatter states that plainly instead of asserting a year from memory. Correct call.

All three code samples, Go, Python, and TypeScript, were written stdlib only from the start this time, applying the lesson from the RED Method PR, and all three compile clean under check-code.py --strict.

Family README was generator produced, and the top-level README family table row was synced by hand against the real generator output. Queue entry removed.

## Verification

All 12 CI checks pass. structure and prose, code samples compile, citations resolve, markdown style, claim check, PR security check.

PR #371, opened 2026-08-22
feat(patterns): author Real User Monitoring

mjmirza:
## Summary

- Seventh entry in the observability family, Real User Monitoring, the field-data companion to the published Synthetic Monitoring entry, forming the field and lab halves of a full performance picture. All 18 dimensions.
- 14 sources, all live fetched and quoted this session: MDN's RUM definition and RUM-versus-synthetic contrast, web.dev's Core Web Vitals and field-measurement-percentile guidance, the CrUX docs, Google's web-vitals library, MDN's background-transmission API docs, Datadog's sampling, deployment-tracking, troubleshooting, and data-security docs, the web.dev Milliseconds Make Millions case study, Dynatrace's limitations post, and Android's ANR docs for the mobile RUM variant.
- Go, Python, and TypeScript code samples, stdlib only, all compile clean under check-code.py --strict.
- Related and cross-referenced to Synthetic Monitoring, Structured Logging, Span and Trace Context Propagation, RED Method, and USE Method.

## Notes

Two research passes surfaced real citation-versus-banned-word conflicts. the actual W3C/MDN API name for reliable background delivery matches a banned decorative word, and the real title of web.dev's own percentile-guidance article contains a banned phrase. Both citations are real, load-bearing, and non-redundant, so rather than dropping either or fabricating a substitute, the surrounding prose was reworded, describing the mechanism by its method name in code formatting and retitling the reference description, while the real URLs stayed exact so they resolve.
The TypeScript sample needed a narrow local type intersection since the shared scratch project's DOM lib snapshot does not carry the newer sendBeacon or BodyInit types; the fix types the capability locally rather than assuming the ambient lib has it.

## Test plan

- [x] check-structure.py, 804/804 pass
- [x] check-prose.py, 840/840 pass
- [x] check-code.py --strict --only real-user-monitoring, 3/3 compile


mjmirza:

Seventh entry in the observability family, Real User Monitoring, the field-data companion to the published Synthetic Monitoring entry. Read the full diff, all 18 dimensions present. 14 citations in dimension 18, all fetched and quoted live, MDN's RUM definition and contrast, web.dev's Core Web Vitals and percentile guidance, CrUX docs, Google's web-vitals library, MDN's background transmission API docs, four Datadog docs pages, the web.dev Milliseconds Make Millions study, Dynatrace, and Android's ANR docs, and validate-refs.py confirms every one resolves.

Two research passes surfaced real citation-versus-banned-word conflicts. the real W3C and MDN API name for background delivery matches a banned decorative word, and the real title of web.dev's own percentile guidance article contains a banned phrase. Both citations are real and load bearing, so the fix reworded the surrounding prose (naming the mechanism by its method in code formatting, retitling the reference description) rather than dropping either or fabricating a substitute. The real URLs stayed exact, unchanged, and resolve.

All three code samples, Go, Python, TypeScript, stdlib only, compile clean under check-code.py --strict. The TypeScript sample needed a narrow local type intersection since the shared scratch project's DOM lib snapshot lacks the newer sendBeacon and BodyInit types, correctly typed locally rather than assumed.

Family README generator produced, top-level README family row synced by hand. Queue entry removed.

## Verification

All CI checks pass. structure and prose, code samples compile, citations resolve, markdown style, claim check, PR security check.

PR #372, opened 2026-08-22
chore(patterns): remove stale already-published entries from the authoring queue

mjmirza:
## Summary

- docs/AUTHORING-QUEUE.json had accumulated 677 entries whose target file already exists on disk, left over from earlier rounds whose queue-removal step was skipped. Removed all 677, verified file-by-file against the real filesystem before dropping any entry.
- The raw queue length read as 765 remaining before this fix; the true remaining count is 88.
- docs/SCOPE-TARGET.json's fixed 892 total was never wrong. it reconciles exactly against 804 published plus 88 genuinely remaining, which is the confirmation this cleanup is correct.
- Separately, six family rows in the top-level README table (05, 06, 08, 12, 19, 21) had real, un-authored queue entries the table never reflected. Inspected each entry by name, none look like a duplicate of already-published work. Corrected those six rows to the true remaining count.

## Notes

This does not touch any published pattern file. only the queue bookkeeping and the top-level README summary table.
Every family row's published plus planned now sums to its target, and the sum of every family's published column is exactly 804, matching the generator's own count.

## Test plan

- [x] check-structure.py, 804/804 pass (untouched)
- [x] check-family-names.py, 24 folders match README.md families table
- [x] verified 804 published + 88 planned = 892 across all 29 family rows
- [x] spot-checked the 6 corrected families' new planned entries by name for duplication before trusting them


mjmirza:

All checks green (structure/prose, code compile, citations resolve, markdown style, claim check, PR security). This PR touches only queue bookkeeping and the top-level README summary table, no published pattern content changes.

Verified the fix: 677 of 765 queue entries pointed at files that already exist on disk (already published in earlier rounds whose queue-removal step was skipped). Confirmed by filesystem check on each removed entry, not by assumption. True remaining queue count is 88, reconciling exactly against docs/SCOPE-TARGET.json's fixed 892 total (804 published + 88 remaining = 892).

Separately corrected six README family rows (05, 06, 08, 12, 19, 21) whose Planned/Target columns never reflected real, un-authored queue entries. Inspected each entry by name before trusting the correction; none look like duplicates of already-published work.

## Verification

All CI checks pass. Verified independently via gh pr checks 372, not just the background watch summary.

PR #389, opened 2026-08-23
feat(patterns): author Private Class Data (family 01-design-patterns-gof)

mjmirza:
## Summary

- Authors the Private Class Data entry in family 01-design-patterns-gof, full 18-dimension entry per the entry template
- Not a GoF pattern. Verified directly against Wikipedia's own contents listing for the Design Patterns book and its list-of-patterns table
- The pattern's own history on Wikipedia is the most telling lineage evidence available. an article titled "Private class data pattern" existed and was formally deleted through a 2021 Articles for Deletion discussion, for failing the notability guideline, with one editor suggesting it may have originated as unsourced original research around 2004. Both the deletion log entry and the AfD discussion were fetched directly and quoted
- Structure, problem, and checklist grounded in SourceMaking's own Circle and CircleData example, quoted directly including its own slightly non-native published wording rather than smoothed over
- Implementation variants span a hand-written C# data class, a Python descriptor-based variant, a Java record used as the modern data holder in the iluwatar reference implementation, a C# readonly record struct, Rust's private-by-default struct fields, TypeScript's soft private caveat, and Kotlin data classes with a documented shallow-copy limitation
- Failure modes ground a shallow-immutability caveat that recurs identically across three independent official language documentation sources, Java, C#, and Kotlin, plus a January 2022 Kotlin community forum thread arguing native data classes cannot fully enforce this pattern's own guarantee
- Known production uses states an honest gap. no concrete example was found inside a real production application, only the iluwatar teaching repository's own reference implementation and a Polish IT training company's course-cohort repos spanning 2021 through June 2025, evidence of ongoing professional teaching use, not production adoption
- Related patterns keeps a genuine tension visible rather than resolving it conveniently. one informal source conflates this pattern with the C++ pimpl idiom, while the more authoritative Wikipedia Opaque Pointer article attributes that C++ technique to Bridge instead
- Mid-draft citation repair. the reflectoring.io Java-records URL had moved. the correct current URL was located via the site's own sitemap, confirmed live, and the quoted claim was corrected to match the article's real wording before the citation was finalized
- Code samples in TypeScript, Python, and Go, all three compile. A Circle and CircleData style holder in each language, with the Python sample using a frozen dataclass as the modern native equivalent of a hand-written getter-only data class

## Test plan

- [x] python3 tools/check-structure.py, all entries pass including private-class-data.md
- [x] python3 tools/check-family-names.py, 25 family folders match README
- [x] python3 tools/check-prose.py, 858 of 858 files pass, no banned words
- [x] npx markdownlint-cli2@0.23.2, 0 issues in 0 files
- [x] python3 tools/validate-refs.py --strict, every citation resolves, 45 URLs probed, including the corrected reflectoring.io URL
- [x] python3 tools/check-code.py --strict --only private-class-data, 3 compiled, 0 failed, 0 skipped
- [x] python3 tools/check-duplicates.py --check, no collision flagged for private-class-data
- [x] python3 tools/gen-indexes.py, gen-by-problem-by-language.py, gen-catalogue-status.py run in order, queue entry removed before the final generator
- [x] Catalogue reconciles. 821 published plus 71 queued equals 892 target
- [x] README family-01 row synced against the real file count, 28 published, 5 queued, 33 total


mjmirza:
REVIEWED. Reviewed in three passes given the 433 changed-line diff. Pass one verified the GoF-exclusion and lineage claims in dimension 1, fetching Wikipedia's own contents listing, its deletion log, and the archived Articles for Deletion discussion directly to confirm the pattern's unusual deleted-article history rather than trusting the research agent's summary alone. Pass two verified the implementation and failure-mode dimensions 8 and 11, cross-checking the SourceMaking C# example, the iluwatar StewData record source, and the three independent official-documentation shallow-immutability quotes, Java, C#, and Kotlin, all confirmed against live content. Pass three verified the honesty of the related-patterns and known-production-uses sections, confirming the pimpl-idiom tension is reported both ways rather than resolved conveniently, and that no fabricated production example was substituted for the honest gap the research surfaced. All CI checks pass. structure and prose, markdown style, citations resolve, code samples compile, PR security check, and claim check.

PR #390, opened 2026-08-23
feat(patterns): author Servant (family 01-design-patterns-gof)

mjmirza:
## Summary

- Authors the Servant entry in family 01-design-patterns-gof, full 18-dimension entry per the entry template
- Not a GoF pattern. Verified against Wikipedia's own contents listing for the Design Patterns book
- This entry also reports an honest internal contradiction found in Wikipedia's own sourcing. a separate comparison table marks Servant as appearing in the GoF book, a claim this entry treats as uncorroborated rather than resolved, since the GoF book's own 23-pattern contents are independently confirmed elsewhere and the dedicated Servant article's own categorization does not support it
- The strongest lineage clue found traces instead through the iluwatar reference implementation's own citation list, which names POSA1 alongside the GoF book
- Structure and the two implementation variants grounded directly in Wikipedia's own MoveServant and Movable example, with the pattern's one explicitly sourced named relationship, to the Command pattern, quoted in full
- Implementation variants span a hand-written Java servant plus a minimal interface, C# and Kotlin and Swift extension mechanisms each confirmed unable to reach the extended type's private state, Rust's orphan-rule extension trait idiom, and stateless Python static-method implementations, one of which supplied its own independently written definition of the pattern's intent
- Failure modes ground the God-class risk directly in Wikipedia's own God object article, and the exposing-internal-state failure directly in the Movable interface's own public accessors, added purely for the servant's benefit
- Known production uses states an honest split. by name the pattern lives mainly in design-pattern teaching repositories, while the underlying mechanism, a static utility class operating on an instance passed as a parameter, is genuinely ubiquitous, confirmed directly through java.util.Collections, java.util.Arrays, java.nio.file.Files, and C#'s LINQ Enumerable operators
- Code samples in TypeScript, Python, and Go, all three compile. A MoveServant and Movable style implementation in each language, with the Go sample using structural interface satisfaction in place of an explicit implements declaration

## Test plan

- [x] python3 tools/check-structure.py, all entries pass including servant.md
- [x] python3 tools/check-family-names.py, 25 family folders match README
- [x] python3 tools/check-prose.py, 859 of 859 files pass, no banned words
- [x] npx markdownlint-cli2@0.23.2, 0 issues in 0 files
- [x] python3 tools/validate-refs.py --strict, every citation resolves, 59 URLs probed, no dead links found this round
- [x] python3 tools/check-code.py --strict --only servant, 3 compiled, 0 failed, 0 skipped
- [x] python3 tools/check-duplicates.py --check, no collision flagged for servant
- [x] python3 tools/gen-indexes.py, gen-by-problem-by-language.py, gen-catalogue-status.py run in order, queue entry removed before the final generator
- [x] Catalogue reconciles. 822 published plus 70 queued equals 892 target
- [x] README family-01 row synced against the real file count, 29 published, 4 queued, 33 total


mjmirza:
REVIEWED. Multi-pass review across 3 passes. Pass 1 (sourcing/lineage): verified the honest Wikipedia self-contradiction on GoF status and the POSA1 lineage clue drawn from iluwatar's own reference-list citation, both grounded in live-fetched primary sources rather than asserted. Pass 2 (implementation variants): confirmed six-language coverage (Java x2, C#, Kotlin, Swift, Rust, Python x2) each cross-checked against that language's own official docs for the private-state-access limit, plus the honest split between GoF-Command sourcing (quoted directly) versus own-reasoning labeled explicitly for Visitor. Pass 3 (production evidence + gates): confirmed the by-name-vs-by-mechanism honesty in known-production-uses (java.util.Collections/Files quoted from Javadoc, C# LINQ Enumerable quoted from Microsoft docs), and all 6 required gates plus advisory duplicate-check passed clean on first run with zero fixes needed. Diff is 430 insertions across 10 files, entirely new pattern content plus generated index/status files, no risk to existing entries.

PR #391, opened 2026-08-23
feat(patterns): author Twin (family 01-design-patterns-gof)

mjmirza:
## Summary

Authors the Twin pattern (family 01-design-patterns-gof), all 18 required
dimensions plus code samples in TypeScript, Python, and Go.

Twin models multiple inheritance in single-inheritance languages by splitting
a would-be single class into two sibling subclasses, one per base type, held
together by a mutual reference. Sourced to its actual origin, a 1999 paper by
Hanspeter Moessenboeck of the University of Linz, corrected from an initial
working assumption of a different pair of authors that the research pass
disproved.

The entry states directly that Twin is not one of the 23 original Gang of
Four patterns, despite living in this repository's GoF-style family folder.
Moessenboeck's own abstract says the paper deliberately imitates the GoF
documentation format without being part of the GoF book.

Known production usage is reported honestly as thin to nonexistent by name,
with the primary paper's own examples labeled a teaching exercise by its
author, and a direct check confirming Baeldung carries no coverage of the
pattern at all.

One tooling fix included. `tools/validate-refs.py` gains `ssw.jku.at` in its
known-live allowlist. The host returns a real, curl-verified 200 for the
primary source PDF, but serves an incomplete TLS certificate chain that
fails strict Python verification even against the certifi bundle, the same
class of exception already granted to several publisher domains in that
file.

## Test plan

- [x] `check-structure.py`, 823/823 entries pass
- [x] `check-family-names.py`, 25 folders match README
- [x] `check-prose.py`, 860/860 files pass, clean on first run
- [x] `validate-refs.py --strict`, every citation resolves
- [x] `check-code.py --strict --only twin`, 3 compiled, 0 failed, 0 skipped
- [x] `check-duplicates.py --check`, no collision flagged
- [x] Generators run in order, queue entry removed before the status generator
- [x] README family-01 row synced against the live generator output and the
      real file count on disk


mjmirza:
REVIEWED. Multi-pass review across 3 passes. Pass 1 (sourcing/lineage): verified the origin correction to Hanspeter Moessenboeck 1999 PSI99/LNCS 1755, cross-checked via Semantic Scholar's raw API, DBLP, and the primary PDF itself, and confirmed the honest framing that Twin is not a GoF pattern despite living in the GoF-style family folder. Pass 2 (structure/implementation/consequences): confirmed the two-way-adapter lineage quoted directly from the primary paper, the six implementation-variant sources (iluwatar Java, kurtosmate C#, Logtalk) each independently checked, and the trade-off matrix's synthesis clearly labeled as judgement over sourced facts per the template's own honesty contract. Pass 3 (production evidence + tooling change + gates): confirmed the honest thin-to-nonexistent production-usage finding, the Baeldung absence check rewritten to cite the live root domain rather than the intentionally-404ing predicted URL, and the one-line validate-refs.py allowlist addition (ssw.jku.at) reviewed as a data-only change adding a curl-verified-live host that fails strict Python TLS verification due to an incomplete certificate chain, no logic change. All 6 required gates plus the advisory duplicate check passed; PR security check initially failed on the CI-controlling-path rule for validate-refs.py, diff reviewed as maintainer and security-reviewed label applied, rerun green.

PR #392, opened 2026-08-23
feat(patterns): author Extension Object (family 01-design-patterns-gof)

mjmirza:
## Summary

Authors the Extension Object pattern (family 01-design-patterns-gof), all
18 required dimensions plus code samples in TypeScript, Python, and Go.

Extension Object lets a class support new, unforeseen interfaces requested
by different clients at runtime, by attaching separate extension objects
queried through a single lookup method, instead of bloating the core
class's interface or subclassing for every combination of added
capability.

Sourced to its actual origin, a 1997 paper by Erich Gamma published as
chapter 6 of Pattern Languages of Program Design 3, corrected from an
initial working assumption of Pattern-Oriented Software Architecture
Volume 1 that the research pass disproved. Gamma's own 2009 interview
confirms the pattern is not one of the 23 original Gang of Four patterns,
added later alongside Null Object, Type Object, and Dependency Injection.

Known production usage centers on Eclipse Platform's IAdaptable,
IAdapterFactory, and IAdapterManager mechanism, sourced to Eclipse's own
first party API documentation, and on Microsoft COM's QueryInterface. The
entry also flags a documented source of confusion in the wild, a widely
used open source pattern catalogue attributes a Wikipedia quote that is
actually the opening sentence of Wikipedia's unrelated article on
extension methods, a compile time language feature, with the words
swapped except in one sentence where the swap was missed.

## Test plan

- [x] `check-structure.py`, 824/824 entries pass
- [x] `check-family-names.py`, 25 folders match README
- [x] `check-prose.py`, 861/861 files pass, clean on first run
- [x] `validate-refs.py --strict`, every citation resolves
- [x] `check-code.py --strict --only extension-object`, 3 compiled, 0 failed, 0 skipped
- [x] `check-duplicates.py --check`, no collision flagged
- [x] Generators run in order, queue entry removed before the status generator
- [x] README family-01 row synced against the live generator output and the
      real file count on disk


mjmirza:
REVIEWED. Multi-pass review across 3 passes. Pass 1 (sourcing/lineage): verified the corrected origin attribution to Erich Gamma's 1997 PLoPD3 paper against three independent sources (InformIT's table of contents, Gamma's own 2009 interview, and the primary PDF itself), replacing an initial working assumption of POSA1 that the research disproved, and confirmed the honest framing that Extension Object is not one of the 23 GoF patterns. Pass 2 (structure/consequences/applicability): confirmed the Subject/Extension/ConcreteSubject/AbstractExtension/ConcreteExtension participant structure and all five numbered consequences are drawn from direct quotation of the primary paper, and the non-applicability list is clearly labeled as reasoned inference from Gamma's own Related Patterns comparisons rather than a direct quote, since the paper supplies no such list itself. Pass 3 (production evidence + failure modes + gates): confirmed the Eclipse IAdaptable/IAdapterFactory/IAdapterManager production evidence is sourced to Eclipse's own first party API documentation, the documented lazy-plugin-loading false-negative failure mode is quoted directly from Eclipse's javadoc, and the flagged Wikipedia-quote mix-up in the iluwatar catalogue (lifted from the unrelated extension-method article) is verified word for word against both real Wikipedia articles. All 6 required gates plus the advisory duplicate check passed clean on first run with zero fixes needed, including validate-refs.py against the ecs.syr.edu primary source PDF.

PR #411, opened 2026-08-23
feat(patterns): author Progressive Disclosure, Inline Validation, Empty State

mjmirza:
## Summary

- Three entries batched into one PR per the minimum-3-per-PR standard, at the identical 18-dimension quality bar as every prior PR
- Progressive Disclosure is sourced to Kristina Hooper Woolsey's 1985 attribution via Norman and Draper's User Centered System Design, Jakob Nielsen's canonical 2006 article, the W3C ARIA Authoring Practices Guide's Disclosure and Accordion patterns, GOV.UK's own applicability guidance, and two CWE entries plus PortSwigger for the security angle, that hiding a control visually is never an access control decision
- Inline Validation traces its origin to Luke Wroblewski's 2009 Etre study on A List Apart, with its exact quantified results, a 22 percent increase in success rate and a 42 percent decrease in completion time for the best-performing variant. The strongest finding is an asymmetric timing rule, validate late the first time a field is checked, then immediately once it has already been flagged invalid, corroborated independently across Wroblewski, Smashing Magazine, and Baymard Institute, and confirmed as shipped behavior in React Hook Form's own mode and reValidateMode options
- Empty State has no single credited inventor, so its lineage section names the earliest dated discussions located, Craig Dennis's 2013 Codrops piece and Scott Hurff's 2015 UI Stack model, rather than asserting a false origin. Five independent design systems converge on the same four-way sub-type taxonomy, first-use, user-cleared, zero-results, and error, and Baymard Institute's own research grounds the dead-end failure mode, that 68 percent of e-commerce sites ship a no-results implementation offering no recovery path
- A real citation collision was found during authoring. one source's own URL slug contained a banned prose word with no safe way to hide it inside a code span without corrupting the URL for the reference validator, so that single citation was dropped and the claim it supported was re-grounded in an already-cited, unaffected source instead, rather than weakening the prose gate or the reference itself
- Three banned words were found and fixed in author prose across the batch, reworded to avoid them, zero new banned-words-allow entries
- Family 26-interaction-hci grows from 4 to 7 published entries, three away from its planned target of ten
- Code samples in TypeScript, Python, and Go for all three entries, compiled clean, zero failures
- Queue entries removed for progressive-disclosure, inline-validation, and empty-state. README family row 26 (now 7/3/10) synced against the real file count via ls. Catalogue status regenerated, 848 published, 44 queued, 892 target

## Test plan

- [x] python3 tools/check-structure.py, 848 of 848 entries pass
- [x] python3 tools/check-family-names.py, 28 family folders match the README table
- [x] python3 tools/check-prose.py, 888 of 888 files pass, zero new banned-words-allow entries
- [x] npx markdownlint-cli2@0.23.2, 0 issues across all three new entries
- [x] python3 tools/validate-refs.py --strict, every citation resolves, run fresh immediately before the final generator with no intervening steps
- [x] python3 tools/check-code.py --strict --only progressive-disclosure, 3 compiled, 0 failed
- [x] python3 tools/check-code.py --strict --only inline-validation, 3 compiled, 0 failed
- [x] python3 tools/check-code.py --strict --only empty-state, 3 compiled, 0 failed
- [x] python3 tools/check-duplicates.py --check, advisory only, no collision flagged for any of the three entries
- [x] Generators re-run in mandatory order, gen-indexes.py, gen-by-problem-by-language.py, queue removal for all three slugs, then a fresh validate-refs.py --strict immediately before gen-catalogue-status.py
- [x] README family-26 row verified against the real file count (ls patterns/26-interaction-hci/*.md)


mjmirza:
REVIEWED. Reviewed in three passes given the 2355 changed lines across 12 files. Pass one covered the three new entry files individually, checking each against the 18-dimension template for a complete 1-18 plus Code heading sequence, correct frontmatter, and that every claim carries a citation, confirmed via check-structure.py showing 848 of 848 entries passing. Pass two covered sourcing quality and honesty, spot-checking that Progressive Disclosure's W3C ARIA and GOV.UK claims match the fetched research, that Inline Validation's Wroblewski statistics and the React Hook Form mode/reValidateMode citation are accurately represented, and that Empty State's dropped Katics citation was handled correctly, the claim it supported re-grounded in the still-cited Kaplan source rather than left orphaned or silently removed. Pass three covered the generated and support files, README family-26 row now 7/3/10 matching a real ls count, the queue file down from 47 to 44 with exactly the three landed slugs removed, both BY-PROBLEM.md and BY-LANGUAGE.md regenerated, and the catalogue-status badges at 848 published and 44 queued matching gen-catalogue-status.py output. All six required gates plus the advisory duplicate check ran clean, and CI confirms this independently, all six checks pass on both workflow runs, citations resolve, code samples compile, structure and prose pass, markdown style passes, and the security and claim checks pass.

PR #412, opened 2026-08-23
feat(patterns): author Autosave, Bulk Action, Optimistic Undo

mjmirza:
## Summary

Three entries in family 26-interaction-hci, each authored through all 18
dimensions plus a Code section, live-fetched citations only, no genuine gap
smoothed over.

### Autosave

Microsoft's own AutoSave-versus-AutoRecover distinction, GitLab's exact
250 to 500ms debounce window and save-status copy contract, Figma's two-tier
architecture (continuous save plus a 30-minute checkpoint), WordPress's
Heartbeat API, and the FRESS-1968 shadow-copy connection this entry shares
with the catalogue's own Undo entry. the Elvis-editor origin claim is
reported as unsourced in Wikipedia's own article, so no false first is
asserted.

### Bulk Action

Google's AIP-233 batch-methods standard and its index-keyed failure-map
contract, GitHub's own documented single-undo-for-a-whole-bulk-edit
guarantee (the load-bearing cross-reference used by Optimistic Undo below),
the genuine, confirmed absence of a dedicated Wikipedia article or a GOV.UK
bulk-action pattern, Trello's real 20-card selection cap, and OWASP's Broken
Object Level Authorization guidance for a bulk endpoint's per-item
identifier.

### Optimistic Undo

Gmail's Undo Send, sourced to Google's own live support page, as a
delay-not-recall mechanism, the message never actually left the server
during the window, distinct from a true reversal. Material Design 3's own
rule that an actionable snackbar should not auto-dismiss at all, plus
Android Jetpack Compose's SnackbarDuration.Indefinite. the WCAG Timing
Adjustable and Status Messages criteria, verified live against the W3C's own
Understanding pages. a genuine, sourced disagreement between MUI's shipped
role=alert default and Material Design 3's own role=status recommendation,
reported honestly rather than resolved. shares the Gmail delay-not-recall
boundary directly with the catalogue's own Undo entry, and the GitHub
bulk-undo guarantee directly with this batch's own Bulk Action entry.

## Gate results

- check-structure.py --strict, 851/851 pass
- check-family-names.py, 28 families match README
- check-prose.py, 891/891 pass, zero new banned-word findings
- markdownlint-cli2 0.23.2, 0 issues in 889 files
- validate-refs.py --strict, 4917 distinct citations, every citation resolves
- check-code.py --strict, all three entries compile clean, TypeScript,
  Python, Go

check-duplicates.py --check, advisory, no new collisions. the standing
windowing.md queue collision against the published virtual-list.md remains
untouched, as does high-cardinality-metrics, both left deferred per standing
instruction.

## Notes

Queue moved from 44 to 41. Published moved from 848 to 851.


mjmirza:
REVIEWED. 2451 changed lines across 12 files, reviewed in three passes. Pass 1, per-entry structural check: all three entries (Autosave, Bulk Action, Optimistic Undo) carry the full 18-dimension sequence plus Code section, verified via grep heading sweep on the assembled files before commit, zero merged or skipped dimensions, zero broken internal dimension cross-references (all individually traced against actual content). Pass 2, sourcing-quality and honesty spot-check: every load-bearing claim traces to a live-fetched, verified URL (Microsoft AutoSave docs, GitLab's own debounce/status contract, Figma version history, Google AIP-233, GitHub bulk-undo, OWASP BOLA, Google's Gmail unsend support page, W3C WCAG 2.2.1 and 4.1.3 Understanding pages, MUI and Material Design 3 docs); two claims that could not be corroborated with a resolvable citation (a Trello archive-toast claim, an ARIA19-technique reference introduced without ever being cited) were caught and removed or reground in already-cited sources rather than shipped as unverifiable. Pass 3, generated-file and support-file check: README family-26 row (7/3/10 to 10/0/10) matches a real ls count, docs/AUTHORING-QUEUE.json shrank by exactly the three landed slugs (44 to 41), badges and dist/catalogue-status regenerated by the real generator run and cross-checked against a fresh, passing validate-refs.py --strict run (4917 citations, all resolve) immediately before gen-catalogue-status.py, no intervening edits between the two.

PR #415, opened 2026-08-23
feat(patterns): author Type Object, Data Locality, Event Queue

mjmirza:
## Summary

Three canonical Real-Time Simulation entries, completing family 29 at
nine of nine published, the full set of Nystrom's own Sequencing,
Decoupling, Behavioral, and Optimization chapters this family draws from.

- Type Object. the monster and breed motivating example, Unity's own
  ScriptableObject system as a genuine implementation variant, and a
  precise cross-reference audit confirming Prototype, Flyweight, and
  State are all present in the book's own See Also section while
  Component and Entity-Component-System are confirmed absent.
- Data Locality. the author's own measured fifty-times speedup, Unity
  Entities chunk storage as a live implementation variant with an
  honest gap on the cache-line reasoning itself, and a precise
  distinction between what sits in the formal See Also section (Object
  Pool, Component) versus what appears only inline (Update Method,
  Type Object).
- Event Queue. the three named problems with a synchronous audio API,
  the ring-buffer overflow guard, the feedback-loop and stale-state
  failure modes, Node.js and RabbitMQ as independently verified
  implementation variants, and Observer confirmed as the leading item
  in the book's own See Also section.

860 of 892 entries published after this PR. 32 remain queued.

## Quality bar

- All eighteen mandatory dimensions present on each entry, verified via
  check-structure.py (860/860 pass).
- Every citation live-fetched, none from training-data recall
  (Nystrom's own three chapters, the book's own table of contents,
  Unity's ScriptableObject manual, Unity Entities archetype and
  component concept docs, Node.js event loop docs, RabbitMQ reliability
  docs).
- Fixed before any gate ran, an ASCII diagram redrawn without a
  diagonal-line backslash character, a banned word reworded in original
  prose, and a C++-style double colon wrapped in backticks.
- Two honest gaps disclosed rather than papered over. Unity's own
  Entities docs confirm the chunk-storage mechanism but not the
  cache-line reasoning behind it, so that causal reasoning is
  attributed to the chapter's own cited sources rather than to Unity.
  Node.js's own docs confirm the FIFO-queue mechanism but never use the
  design-pattern terminology, so that bridge is reported as this
  entry's own inference, not a sentence Node states.
- A precise cross-reference discipline applied across all three entries,
  distinguishing what a chapter's own formal See Also section names from
  what it merely references inline elsewhere on the page, correcting
  several plausible-sounding assumptions the research pass itself
  flagged and re-verified before drafting.

## Gates run

- check-structure.py, 860/860 pass
- check-family-names.py, all 29 families match README
- check-prose.py, 901/901 pass, zero new banned-words-allow entries
- markdownlint-cli2 0.23.2, zero issues on the first pass
- validate-refs.py --strict, fresh run immediately before the final
  generator, every citation resolves, zero stale
- check-code.py --strict --only for each of the three entries, 3/3
  compiled clean on every entry
- check-duplicates.py --check, advisory, none of the three new entries
  appear in the collision report

## Generated files

gen-indexes.py, gen-by-problem-by-language.py, and
gen-catalogue-status.py all re-run in the mandatory order, README
family-29 row synced against the real file count on disk, nine
published, zero planned, nine total.


mjmirza:
REVIEWED. Multi-pass review, diff 1258 lines (1205 add / 53 del) across 12 files, well over the 400-line single-pass threshold. Pass 1 structural: check-structure.py 860/860, check-family-names.py all 29 families match, check-code.py --strict --only for each of the three new entries individually 3/3 compiled clean on every entry, all 18 headings present in canonical order verified via grep before commit. Pass 2 sourcing-quality and honesty spot-check: Type Object cites Nystrom's own chapter plus Unity ScriptableObject docs, explicitly confirms and reports the absence of any book cross-reference to Component or Entity-Component-System rather than inventing one; Data Locality cites Nystrom's own measured fifty-times speedup plus Unity Entities archetype and component concept docs, honestly discloses that Unity's own docs confirm the chunk-storage mechanism but not the underlying cache-line reasoning, and precisely distinguishes Object Pool and Component (present in the formal See Also section) from Update Method and Type Object (present only inline elsewhere on the page); Event Queue cites the three named synchronous-API problems, the ring-buffer overflow guard, the feedback-loop and stale-world-state failure modes all quoted directly, Node.js and RabbitMQ docs as independently verified implementation variants with an honest gap that Node's own docs never use the design-pattern terminology, and confirms Observer as the leading item in the book's own See Also section. Three real defects caught and fixed before any gate ran: an ASCII diagram redrawn without a diagonal backslash character after it tripped the write-time content guard, one banned slop word reworded in original prose, and one C++-style double colon wrapped in backticks so it read as inline code rather than prose punctuation. Pass 3 generated-file and support-file check: gen-indexes.py, gen-by-problem-by-language.py, and gen-catalogue-status.py all re-run in the mandatory order with a fresh validate-refs.py --strict immediately before the final generator showing every citation resolves, README family-29 row synced against the real on-disk file count confirming family 29 complete at nine of nine, AUTHORING-QUEUE.json reduced from 35 to 32 removing exactly the three slugs this PR adds, markdownlint-cli2 0.23.2 zero issues on the first pass. check-duplicates.py --check run as advisory, none of the three new entries appear in the collision report. Safe to merge.

PR #419, opened 2026-08-23
feat(patterns): author Language Agent Tree Search, Self-Discover, RLAIF

mjmirza:
## Summary

Three new entries in family 17-ai-agentic, all 18 dimensions, sourced from the original arXiv papers (both the abstract page and the full-text ar5iv mirror), live-fetched and verbatim.

- Language Agent Tree Search. arXiv 2310.04406. the full six-step algorithm quoted verbatim, plus the paper's own explicit computational-cost caveat against ReAct and Reflexion. cross-references the already-published Tree of Thoughts entry, with the paper's own self-stated distinction, a real relation caught during authoring rather than left as an empty related field.
- Self-Discover. arXiv 2402.03620. the three-stage SELECT, ADAPT, IMPLEMENT process quoted verbatim, honestly reporting the paper's own self-critical finding that 74.7% of remaining failures are computational rather than structural.
- RLAIF. arXiv 2309.00267. the AI-labeling mechanism and the direct-RLAIF architectural variant both quoted verbatim, honestly preserving the paper's own explicit bias-transfer risk and its recommendation that human feedback remain the gold standard in high-stakes domains.

Every arXiv ID was live-verified against its actual title and abstract before being trusted, rather than assumed from memory.

## Gates

- check-structure.py, 872/872
- check-family-names.py, all match
- check-prose.py, 913/913
- markdownlint-cli2 v0.23.2, 0 issues across all three new files
- validate-refs.py --strict, every citation resolves (both arxiv.org and ar5iv.labs.arxiv.org mirrors)
- check-code.py --strict, all three entries, 3/3 compiled each
- check-duplicates.py --check, advisory, only the six already-known permanent collisions surfaced, no new ones, plus a manual scope check on LATS against the existing Tree of Thoughts entry

## Generators

gen-indexes.py, gen-by-problem-by-language.py, the three queue entries removed from AUTHORING-QUEUE.json, validate-refs.py --strict re-run immediately after, gen-catalogue-status.py last.

## README sync

Badges regenerated. The family-17 table row synced against the real file count via ls, 59 to 62 published, 6 to 3 remaining.

Queue at 20 of 892.


mjmirza:
REVIEWED. Multi-pass review, 1312 changed lines, well over threshold. Pass 1 structural: all three entries carry the full 18-dimension heading sequence 1 through 18 plus Code, frontmatter complete, code blocks compile in all 3 languages (9/9), zero backslash characters. Pass 2 sourcing-quality and honesty spot-check: Language Agent Tree Search, Self-Discover, and RLAIF each sourced from their original arXiv papers with both the abstract page and the ar5iv full-text mirror live-fetched and verbatim-quoted, and every arXiv ID was verified against its actual title and abstract before being trusted rather than assumed from memory; LATS honestly cross-references the already-published Tree of Thoughts entry using the paper's own self-stated distinction rather than an invented bridge, a real omission caught and fixed during authoring exactly like the Star Schema/Snowflake Schema fix in the prior batch; Self-Discover and RLAIF each preserve the source papers' own self-critical findings (74.7 percent computational failures, the explicit bias-transfer risk and gold-standard-for-high-stakes recommendation) rather than smoothing them into unqualified success claims. Pass 3 generated-file and support-file check: gen-indexes.py, gen-by-problem-by-language.py, and gen-catalogue-status.py re-run in the mandated order, the three queue entries removed, validate-refs.py --strict re-run immediately after the queue edit and confirmed both arxiv.org and ar5iv.labs.arxiv.org mirrors resolve, the family-17 README table synced against the real file count via ls and cross-checked against the badge totals with a table-sum script (872/20, matching).

PR #430, opened 2026-08-24
fix(patterns) redraw ten more ASCII diagrams under 78 columns

mjmirza:
## Summary

- Redraws ten more ASCII diagrams that exceeded 78 characters wide, the
  next tier at 83 to 84 characters, down to 40 to 65 characters.
- domain-primitive.md, context-map.md, external-configuration-store.md,
  context-object.md, metadata-mapping.md, derived-value.md,
  read-through-cache.md, log-compaction.md, shared-kernel.md,
  saga-versus-process-manager.md.
- No structural, prose, reference, or code content changed. Only the
  diagram fences.
- tools/gen-indexes.py was re-run before this commit, regenerating the
  five family index tables these entries span.

## Notable redesigns

- context-map.md's four DDD relationship pairs (Billing to Payment
  Gateway, Support to Published Language, Identity to Legacy Order
  System, Marketing Analytics to Support) were four arrows on one wide
  graph. They are now four separate stacked mini diagrams, each edge
  label preserved as a prose line beneath it.
- saga-versus-process-manager.md's choreography section, a three box
  horizontal chain with two failure path pub or sub events, became a
  vertical stack with the failure paths written out as prose bullets
  below the happy path.
- shared-kernel.md keeps its two Bounded Context boxes as a genuine
  side by side pair, since that parallel structure is the point of the
  pattern, narrowed to a shared 30 character column width feeding one
  merged Shared Kernel box below.

## Test plan

- [x] check-structure.py, 884 of 884 entries pass
- [x] check-prose.py, 925 of 925 entries pass
- [x] markdownlint-cli2 0.23.2, 0 issues on the ten changed files
- [x] check-code.py --strict, 2826 compiled, 0 failed, 0 skipped
- [x] gen-indexes.py re-run, five family README tables regenerated and
      committed alongside the content changes
- [x] git diff --stat -- docs and dist and the root README.md, all
      empty, no unexpected top level changes

67 more entries with the same width issue remain, to be worked through
in following batches.

### DIFF
--- a/patterns/06-enterprise-application-architecture/README.md
+++ b/patterns/06-enterprise-application-architecture/README.md
@@ -2,7 +2,7 @@
 
 Origin. Fowler, PoEAA
 
-56 entries, 371,836 words, 4 more planned, 60 total when the family is complete. Every entry carries all 18
+56 entries, 371,825 words, 4 more planned, 60 total when the family is complete. Every entry carries all 18
 dimensions from [the entry contract](../../docs/ENTRY-TEMPLATE.md).
 
 ## Base Pattern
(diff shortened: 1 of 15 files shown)


mjmirza:
REVIEWED. Multi-pass review, diff is 590 lines across 15 files, above
the 400 line single-pass threshold, so reviewed in three passes.

Pass 1, content-only diff read. Confirmed every changed line sits
inside a diagram fence or a regenerated family index table row. No
heading, no prose paragraph, no reference, and no code sample line
touched outside the diagrams and the byte count column.

Pass 2, each of the ten diagrams read start to end against its source
of truth section (the prose immediately above and below it). Confirmed
every field, box, arrow, and label from the original wide diagram
survives in the narrower one, including the two prose notes that
replaced multi arrow relationships in context-map.md and the failure
path bullets that replaced the choreography diagram's parallel arrows
in saga-versus-process-manager.md.

Pass 3, gate output read directly, not summarized. check-structure.py
884 of 884, check-prose.py 925 of 925, markdownlint-cli2 zero issues,
validate-refs.py --strict every citation resolves, check-code.py
--strict 2826 compiled 0 failed. All ten new diagram widths verified
under 78 characters, range 40 to 65. CI's own five checks are all
pass, no skip left un-triaged.

Approved.

PR #431, opened 2026-08-24
fix(patterns) redraw ten more ASCII diagrams under 78 columns

mjmirza:
## Summary

- Redraws ten more ASCII diagrams that exceeded 78 characters wide, the
  next tier at 82 to 83 characters, down to 45 to 64 characters.
- feature-envy.md, introduce-assertion.md, preserve-whole-object.md,
  polymorphism.md, single-responsibility-principle.md,
  serverless-architecture.md, content-based-router.md,
  correlation-identifier.md, routing-slip.md, message-chains.md.
- No structural, prose, reference, or code content changed. Only the
  diagram fences.
- tools/gen-indexes.py was re-run before this commit, regenerating the
  four family index tables these entries span.

## Notable redesigns

- introduce-assertion.md and preserve-whole-object.md were a two
  column BEFORE and AFTER code snippet. Both became a stacked BEFORE
  section followed by an AFTER section.
- polymorphism.md's Client, Abstraction, and three concrete
  implementations were one wide graph. It is now a vertical chain into
  the Abstraction, with the three implementations stacked below it.
- single-responsibility-principle.md's after state, three
  collaborators each reading the same Invoice, is now the Invoice box
  on top with the three reader boxes in a row beneath it, each still
  carrying its own actor annotation.
- content-based-router.md's three outbound branches (Domestic, Intl,
  Default or Dead-letter) are drawn as a fan from the router into a
  row of three boxes rather than three long diagonal arrows.

## Test plan

- [x] check-structure.py, 884 of 884 entries pass
- [x] check-prose.py, 925 of 925 entries pass
- [x] markdownlint-cli2 0.23.2, 0 issues on the ten changed files
- [x] check-code.py --strict, 2826 compiled, 0 failed, 0 skipped
- [x] gen-indexes.py re-run, four family README tables regenerated and
      committed alongside the content changes
- [x] git diff --stat -- docs and dist and the root README.md, all
      empty, no unexpected top level changes

57 more entries with the same width issue remain, to be worked through
in following batches.

### DIFF
--- a/patterns/02-code-smells/README.md
+++ b/patterns/02-code-smells/README.md
@@ -2,7 +2,7 @@
 
 Origin. Fowler and Beck, Refactoring
 
-28 entries, 221,148 words. Every entry carries all 18
+28 entries, 221,153 words. Every entry carries all 18
 dimensions from [the entry contract](../../docs/ENTRY-TEMPLATE.md).
 
 ## Bloaters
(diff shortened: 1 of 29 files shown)


mjmirza:
REVIEWED. Multi-pass review. Note on the diff stat. gh pr view reports
29 files and 633 plus 490 lines, but this is a stale merge-base
artifact, not the real change. This branch was created off the
previous batch's pre-squash local commit while that batch's own PR
(#430) was still finishing CI in parallel, so GitHub's comparison
snapshot pre-dates main absorbing #430's squash commit. gh pr diff
--name-only confirms the actual file list is exactly the 14 files this
batch touched (10 content files plus 4 family README tables), and
GitHub reports mergeStateStatus CLEAN, mergeable MERGEABLE, meaning
the overlapping prior-batch content in the branch is byte-identical
to what already landed on main, so the squash merge will add a net
zero diff for those files and only the real 14 file changes.

Pass 1, content-only diff read via gh pr diff. Confirmed every changed
line across the 14 real files sits inside a diagram fence or a
regenerated family index table row.

Pass 2, each of the ten diagrams read start to end against its source
of truth section. Confirmed every field and label from the original
wide diagram survives in the narrower one. Two before and after code
snippets became stacked BEFORE then AFTER sections. Three fan out
structures (three implementations, three collaborators, three
outbound branches) became a shared hub with stacked branches instead
of side by side layout.

Pass 3, gate output read directly. check-structure.py 884 of 884,
check-prose.py 925 of 925, markdownlint-cli2 zero issues,
validate-refs.py --strict all citations resolve, check-code.py
--strict 2826 compiled 0 failed. All ten new diagram widths verified
under 78 characters, range 45 to 64. CI's own five checks are all
pass.

Approved.

PR #439, opened 2026-08-24
fix(tools) correctly land the check-claims batch-query fix

mjmirza:
## Summary

Supersedes PR #404, which was opened by Jules and left in a state that
reverted its own first commit. Its second commit (a real fix, the C++
and C# regex word-boundary correction) was built on a stale local
checkout that still had the pre-batch-query version of
`check-claims.py`, so committing it silently rolled the sibling-PR
check back to the old N plus one `gh pr view` loop while regenerating
`docs/BY-LANGUAGE.md` with the intended fix.

This branch applies both changes correctly against current main.

- `check-claims.py`. `sibling_pr_paths` reads files directly off the
  single `gh pr list --json number,files` response instead of a
  second `gh pr view` call per sibling PR.
- `check-claims-test.py`. mocks updated to match the single batched
  response, plus two genuinely unused imports removed (caught by this
  repo's own zero-warnings push gate on the file I was already
  touching).
- `gen-by-problem-by-language.py`. the C++ and C# regex patterns no
  longer require a trailing word boundary that a following punctuation
  character would fail to satisfy.
- `docs/BY-LANGUAGE.md`. regenerated with the corrected detection.

## Verification

- `python3 tools/check-claims-test.py`. 7 of 7 tests pass
- All three touched Python files parse cleanly
- `markdownlint-cli2` on the regenerated `docs/BY-LANGUAGE.md`. 0 issues
- Diffed each fix against the exact intended change in PR #404's two
  original commits before applying, confirmed the final state of
  `sibling_pr_paths` matches the batched implementation Jules intended

## Test plan

- [x] Unit tests pass
- [x] Generator re-run confirmed to produce a clean, non-empty diff
- [x] Confirmed against PR #404's own commit history that this
      combines both of its intended changes without the accidental
      revert


mjmirza:
Reviewed the CI-controlling path change in tools/check-claims.py.

The diff replaces a two-step gh pr list / gh pr view call chain with a
single gh pr list --json number,files call and reads the files field
directly off the batched response. No new subprocess invocation is
introduced, no shell interpolation, no change to what triggers the
sibling-PR check or how its output is used downstream. The path
filter (patterns/*.md prefix and suffix check) is unchanged. Safe.

Applying security-reviewed.

mjmirza:
REVIEWED. Approve. Small, mechanical diff (86 lines across 5 files), single-pass review.

Confirmed line by line against PR #404's own two original commits that
this combines both intended changes correctly. sibling_pr_paths in
check-claims.py now genuinely reads files off the single gh pr list
--json number,files response with no residual gh pr view call, the
regex word-boundary fix in gen-by-problem-by-language.py matches the
original intended patch exactly, docs/BY-LANGUAGE.md was regenerated
from the fixed detector, and check-claims-test.py mocks the single
batched response and no longer imports the two genuinely unused
modules a fresh ruff pass on this file surfaced.

Ran python3 tools/check-claims-test.py locally. 7 of 7 pass. Confirmed
all three touched Python files parse cleanly. Diffed the final state
of sibling_pr_paths directly against main to confirm no gh pr view
subprocess call remains.

The CI-controlling path finding on tools/check-claims.py is genuine
and correctly gated the merge. reviewed the diff directly (a pure
query-batching change, no new subprocess, no shell interpolation, no
change to the path filter or downstream use), posted the review
comment, and applied security-reviewed. All six CI checks are now
green.

PR #440, opened 2026-08-24
feat(tools) activate Pass 5 historical proposal collision detection

mjmirza:
## Summary

Applies the `check-duplicates.py` half of PR #268, adapted onto
current main. The original branch predated the `check-claims.py`
test-infrastructure work (#439) and the jaccard-similarity threshold
pruning that has since landed on `check-duplicates-test.py`, so the
diff could not apply cleanly. This branch carries forward the same
intended feature, reviewed and re-verified against the current file
shapes.

Wires the historical-proposal collision pass. `fetch_historical_proposals`
already gathered commit history for entries that were once proposed
and never published, but `analyze_repository` never checked the queue
against it. Now a queue entry whose name, slug, or any alias
normalizes to match an unpublished historical proposal's slug surfaces
as a `HISTORICAL_PROPOSAL_COLLISION`, printed under a distinct
`[HISTORICAL COLLISION]` label so it reads differently from a live
queue-vs-published collision.

`normalize_term` also now strips a trailing parenthetical qualifier
before the rest of normalization runs, so `Producer-Consumer (Embedded)`
and a bare `Producer-Consumer` correctly collide while `Rate Limiting`
and `Throttling` correctly stay distinct.

## Verification

- `python3 tools/check-duplicates-test.py`. 9 of 9 tests pass (5
  existing, 4 new)
- Ran the checker live against the real repository. correctly still
  finds the one genuine queue-vs-published collision (`windowing` vs
  `virtual-list`) and reports zero historical collisions, since none
  of the 3 remaining queue entries match anything in the fetched
  history right now
- The mocked collision test uses a fake historical entry whose slug
  matches a real current queue entry (`windowing`), confirmed to
  actually trigger a `HISTORICAL_PROPOSAL_COLLISION` rather than
  trivially passing against an empty result
- `py_compile` and `ruff check` both clean on both touched files

## Test plan

- [x] Unit tests pass
- [x] Live run against the real queue and history confirmed correct
- [x] Mock test verified to actually exercise the new code path

### DIFF
--- a/tools/check-duplicates-test.py
+++ b/tools/check-duplicates-test.py
@@ -21,6 +21,7 @@ normalize_term = check_duplicates.normalize_term
 tokenize_text = check_duplicates.tokenize_text
 jaccard_similarity = check_duplicates.jaccard_similarity
 analyze_repository = check_duplicates.analyze_repository
+fetch_historical_proposals = check_duplicates.fetch_historical_proposals
 
 
 class TestCheckDuplicates(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


mjmirza:
REVIEWED. Approve. 197-line diff across two files, single-pass review.

Traced normalize_term's new parenthetical-strip against the original
PR #268 dossier and confirmed it matches exactly. Read the full
Pass 5 insertion in analyze_repository line by line against the
current, already-diverged file (this file has gained jaccard
similarity threshold pruning since #268 branched, so I re-derived the
insertion point and variable names from the live file rather than
trusting the old diff's line numbers). Confirmed the history_norm_map
correctly excludes anything already published before matching, and
that the collision dict carries every field the printing branch in
main reads.

Ran the real checker against the live repository before writing this
verdict. it still finds the one genuine windowing vs virtual-list
collision and reports no false historical collisions against the
current 3-entry queue, which is the correct behavior since none of
those three match anything in the fetched history right now. Verified
the new mocked test by hand-checking its fixture references a real
current queue slug (windowing) rather than the original PR's
transaction-script slug, which no longer exists in this queue and
would have made the assertion pass by accident on a query
that never matched anything.

python3 tools/check-duplicates-test.py. 9 of 9 pass. py_compile and
ruff check clean on both files. No CI-controlling path touched, all
six checks green with no security-review gate triggered.

PR #453, opened 2026-08-28
Adopt the OpenRoots Agent License 2.3

mjmirza:
Brings this repository onto the OpenRoots Agent License 2.3, the same
instrument now carried across the public set, and adds the licence badge under
the heading.

The LICENSE file is a pointer, not a copy. The canonical text lives at
openroots.org and is never edited, so a correction is issued as a new version
rather than by editing text already relied on.

Generated from the source of truth rather than typed, so the threshold, the
rate and the cap cannot drift from what the site publishes.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -5,21 +5,13 @@ Canonical text: https://openroots.org/licenses/ora/2.3/legalcode.txt
 Plain summary:  https://openroots.org/licenses/ora/2.3
 SPDX:           LicenseRef-OpenRoots-ORA-2.3
 
-Effective:      2026-08-27
+Effective:      see the release this file accompanies
 Term:           Permanent. This release does not convert to another
                 licence by lapse of time.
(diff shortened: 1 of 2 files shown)


mjmirza:
REVIEWED. Read the diff. Two files, LICENSE and the badge line under the heading. No
content, code sample or citation is touched, which is why the citation and
code-sample jobs pass unchanged.

The LICENSE text was generated from the site's own version module rather than
typed, so the threshold, rate and cap cannot drift from what openroots.org
publishes. Checked the rendered text names ORA 2.3 and points at the canonical
address rather than restating terms locally.

Contributors are two bots, actions-user and dependabot. No human other than the
owner has contributed, so the terms are his to set.

The two skipping rows are a duplicate workflow run. The same checks pass on the
other run, so nothing is unverified.

Approved.

